PR #171, opened 2022-05-11
170 open windows

### LINKED ISSUE #170: `wb$open()` not working on Windows
This simply doesn't do anything.

```r
library(openxlsx2)
wb_workbook()$add_worksheet(1)$open()
```

Might be able to change the internal code to use `shell.exec()` instead of the `shell()` command written out.  I usually have better success with that.

jmbarbone:
closes #170 

Some things:

* Opening on windows now uses `shell.exec()` which will do the dirty work of figuring out what to use to open the xlsx file
* `interactive` to control `interactive()` check -- maybe someone wants to force open these things
* `wb$open()` now passes directly to `xl_open()` which clones the object so that `wb$path()` is not affected.  This is nice because `wb$save()` overwrites `wb$path` and now the `xl_open()` is more useful because it will only create a temporary file.
* We use some `system()` calls and `system2()` is recommended.  I didn't changed these because I can't test them right now
* first argument to `xl_open()` changes from `file` to `x` to highlight that you can pass either a path or a `wbWorkbook` object
* general cleanliness


R/xl_open.R
```diff
@@ -30,42 +32,61 @@
...
-    system(command = myCommand)
-  } else {
-    warning("Operating system not handled.")
-  }
+  switch(
+    userSystem,
+    Linux = {
+      app <- getOption("openxlsx2.excelApp", chooseExcelApp())
+      myCommand <- paste(app, file, "&", sep = " ")
+      system(command = myCommand)
```
jmbarbone, line 78: @JanMarvin , do you want to test `system2(app, c(file, "&"))`? I believe that should work

JanMarvin: I can check later this week, though I do not expect `system()` to go away. It's one of the deprecations that lingers around forever.

R/xl_open.R
```diff
@@ -3,8 +3,8 @@
 #' @description This function tries to open a Microsoft Excel (xls/xlsx) file or
 #'   an openxlsx Workbook with the proper application, in a portable manner.
```
JanMarvin, line 4: s/openxlsx/openxlsx2/g :)

R/xl_open.R
```diff
@@ -30,42 +33,61 @@
...
-  }
+  switch(
+    userSystem,
+    Linux = {
+      app <- getOption("openxlsx2.excelApp", chooseExcelApp())
+      myCommand <- paste(app, file, "&", sep = " ")
+      system(command = myCommand)
+    },
+    Windows = {
+      shell.exec(file)
```
JanMarvin, line 82: #nolint at the end. This is only available on windows iirc therefore the lintr barks

R/xl_open.R
```diff
@@ -1,23 +1,26 @@
...
-#' the user via a menu (if many are present, otherwise it will
-#' set the only available). Currently searched for apps are
-#' Libreoffice/Openoffice (`soffice` bin), Gnumeric
-#' (`gnumeric`) and Calligra Sheets (`calligrasheets`).
+#'   In Windows it uses `base::shell.exec()` (Windwos only function) to
+#'   determine the appropriate program.
 #'
-#' @param file path to the Excel (xls/xlsx) file or Workbook object.
-#' @usage xl_open(file=NULL)
+#'   In Mac (c) it uses system default handlers, given the file type.
```
JanMarvin, line 9: I'm not sure, I will test later, but it should be possible to open workbooks with non default system handlers. Aka you could force opening the file with Numbers, but I haven't actually tried this.

R/xl_open.R
```diff
@@ -78,19 +100,26 @@ chooseExcelApp <- function() {
 
   prog <- Sys.which(m)
   names(prog) <- names(m)
-  nApps <- length(availProg <- prog["" != prog])
+  availProg <- prog["" != prog]
```
JanMarvin, line 103: I don't get it. What? 😄

JanMarvin (approved):
I do not really have intentions to open xlsx files on Windows with something other than Excel, therefore 👍

R/xl_open.R
JanMarvin: On mac the following works: 
```bash
open -a "Numbers" filepath/filename
```

R/xl_open.R
```diff
@@ -115,7 +113,7 @@ chooseExcelApp <- function() {
     unnprog <- unname(availProg)
     message(sprintf("Setting options(openxlsx2.excelApp = '%s')", unnprog))
     options(openxlsx2.excelApp = unnprog)
-    invisible(unnprog)
+    return(invisible(unnprog))
```
JanMarvin, line 116: needed the return, otherwise it would always end on the stop at the end of the function

JanMarvin:
Rebased and updated doc and merged. Thanks!

PR #175, opened 2022-05-12
Combine writes

JanMarvin:
Work in progress:

* [x] reduce differences between `write_data()` and `write_datatable()`
  * moved code around
  * moved defined names code to `write_data2()`. it should be replaced with another function that exists for just _this very specific_ case
  * it boils down to two chunks that are exclusively in `data` or `datatable`. A few `assert_class()` calls and minor differences how `write_data2()` and `wb_check_overwrite_tables()` are called
  * there were a few potentially dangerous options, returning `invisible(0)` were warnings or errors are required
  * after the changes a single test had to be updated
* [x] combine both in a single function that can be fully tested
* [x] one or both previous functions can become wrappers of the new function
* [x] tested `write_data()` and `write_datatable()` examples with Excel


R/writeData.R
```diff
@@ -160,98 +163,42 @@ write_data <- function(wb,
...
-    new_cols = c(startCol, startCol + nCol - 1L),
-    check_table_header_only = TRUE,
-    error_msg =
-      "Cannot overwrite table headers. Avoid writing over the header row or see wb_get_tables() & wb_remove_tabless() to remove the table object."
-  )
-
-
-
   ## write autoFilter, can only have a single filter per worksheet
   if (withFilter) {
```
JanMarvin, line 192: wb$add_filter()

R/wb_functions.R
```diff
@@ -1163,6 +1163,264 @@ write_data2 <-function(wb, sheet, data, name = NULL,
...
+  }
+  ### End: Only in data --------------------------------------------------------
+
+  overwrite_nrows <- ifelse(data_table, 1L, colNames)
+  check_tab_head_only <- ifelse(data_table, FALSE, TRUE)
+  error_msg <- ifelse(
+    data_table,
+    "Cannot overwrite existing table with another table",
+    "Cannot overwrite table headers. Avoid writing over the header row or see wb_get_tables() & wb_remove_tabless() to remove the table object."
+  )
```
JanMarvin: Should be in a single `if ... else ...` statement

JanMarvin:
This one should be ready now. I've moved code from writeData.R and writeDataTable.R and combined both in a new meta function(write_data_table()). This meta function bundles another meta function (write_data2()), going full inception. Maybe a few lines could be moved to write_data_table() to trim down write_data2() a bit. But we will need to start looking into this once we refactor the write_data code to move a few lines to wbWorkbook (if we still intend to do so) and for possible speed improvements.

Oh and I did not really touch the code and there are still parts I do not really understand (e.g. the -1 part in the table overwrite check) and some parts should be replaced with existing internal functions (like add_filter()).

JanMarvin:
@jmbarbone I hold this for the day if you want to risk looking at this pre merge, otherwise it goes to main tomorrow. Been sitting at my computers already way to long ...

PR #183, opened 2022-05-16
113 sheet funs

### LINKED ISSUE #113: include `current_sheet()`, `next_sheet()` waivers/private fields
I'm thinking about adding some "last action" entry for `wbWorkbook`, but like to discuss this previously.

My rationale is the following:
```R
wb <- wb_workbook()
wb <- wb_add_worksheet("test") # add wb$last$sheet <- "test"
```

## why
We could use `$last` for different things like `sheet` but also `dims`. I was thinking about this, when playing around with borders. Previous functions had options for borders surrounding some area. Right now, we do not now which area was written. If we would store this via `wb$last$dims` we could simply use the dims to build our borders on this.

### current
So all calls following this call can access `wb$last$sheet` to work on this sheet. So calling
```R
writeData(wb, x = mtcars)
# is identical to
writeData(wb, sheet = "test", x = mtcars)
```

### future
Once `writeData` becomes something like `wb_write_data` the pipe would look something like this:
```R
wb <- wb_workbook() |>
  wb_add_worksheet("Sheet 1") |>
  wb_write_data(x = mtcars)
```

The chain would be:
```R
wb <- wb_workbook()$
   add_worksheet("Sheet 1")$
   write_data(x = mtcars)
```

jmbarbone:
resolves #113


tests/testthat/test-class-workbook-utils.R
```diff
@@ -0,0 +1,18 @@
...
+  expect_identical(wb$.__enclos_env__$private$current_sheet, 0L)
+  expect_error(wb$add_worksheet(), NA)
+  expect_identical(wb$.__enclos_env__$private$current_sheet, 1L)
+  expect_error(wb$add_worksheet(), NA)
+  expect_identical(wb$.__enclos_env__$private$current_sheet, 2L)
+})
+
+test_that("waivers work with $add_data()", {
+  # wbWorkbook$undebug("add_data")
+  # debugonce(write_data_table)
```
JanMarvin, line 12: Unintended include?

JanMarvin (approved):
Wonderful! Love it!

PR #189, opened 2022-05-16
[c++/pugi] use parse_ws_pcdata. closes #188

### LINKED ISSUE #188: Issue with shared strings
There seems to be an issue with shared strings. When opening the file from #186 I get some strange warning in Excel. I have a similar warning if I load and open the xlsx file from this issue: https://github.com/ycphs/openxlsx/issues/81 (this one is about styles.xml)

Unfortunately both files have a huge shared strings table. I've tried to find something but had no luck. Might be something our parses does to the file, but I'm quite clues at the moment.

Iirc copying sharedstrings.xml solved this, but if that's the only solution, we'd never be able to write shared strings.

JanMarvin:
Previously pugi would reduce `<si><t xml:space="preserve"> </t>` to this `<si><t xml:space="preserve"/></si>`. This confused a certain spreadsheet software.

### DIFF
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -1282,7 +1282,7 @@ wbWorkbook <- R6::R6Class(
             length(self$sharedStrings),
             attr(self$sharedStrings, "uniqueCount")
           ),
-          #body = stri_join(set_sst(attr(sharedStrings, "text")), collapse = "", sep = " "),
+          #body = stri_join(set_sst(attr(self$sharedStrings, "text")), collapse = "", sep = " "),
           body = stri_join(self$sharedStrings, collapse = "", sep = ""),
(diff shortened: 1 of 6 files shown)


JanMarvin:
The comment above is still valid, though the path to success was different. I've added `pugi::parse_ws_pcdata` to the parse arguments. (Maybe `pugi::parse_ws_pcdata_single would have had the same effect, though I do not really understand the documentation at this point.)

JanMarvin:
Wohoo I can rest in peace! The evil xlsx file that send me on this mission can finally be imported and exported without Excel complaining! 🥳 
```R
wb_load("https://github.com/ycphs/openxlsx/files/4879159/Workbook.Template.new.xlsx")$open()
```

PR #191, opened 2022-05-21
replace tables list object with data frame

JanMarvin:
* avoid wrangling with list attributes
* avoid having references to deleted tables

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,8 @@
 
 ## Internal changes
 
+* Rewrite `wb$tables` to use a data frame approach. This simplifies the code a bit and makes it easier to implement more upcoming changes [191](https://github.com/JanMarvin/openxlsx2/pull/191)
+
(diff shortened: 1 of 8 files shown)


JanMarvin:
This is a first step to clean up `wb$tables` and table interaction. Previously `attr(foo, "bar")` was used to pass table names and table sheets. This led to complicated code where every time something was changed, the corresponding attributes had to be checked and resized (increased or decreased). This is now avoided. In this first attempt, only the original behavior was restored. So when a table is deleted, it is set to `wb$tables$tab_act <- 0` and the name contains a prefix `deleted_`. This should be changed in a subsequent pull request. There is no need to keep the removed reference anymore.

JanMarvin:
It might be fruitful, to check this a few times with Excel. The previous implementation was a bit shaky, this should be an improvement, especially once the deleted sheets are deleted from the wbWorkbook object when they are removed.

PR #193, opened 2022-05-22
[R/get_named_region] add tables to `get_named_region()` 

JanMarvin:
this allows reading tables with `wb_to_df()` and could be used in `wb_validate_named_region()`.

Updated man page, added test and removed default and wbWorkbook method


tests/testthat/test-named_regions.R
```diff
@@ -402,3 +402,29 @@ test_that("Overwrite and delete named regions", {
...
+    name = "iris",
+    rows = seq_len(nrow(iris) + 1),
+    cols = seq_along(iris)
+  )
+
+  expect_equal(c("iris", "iris_tab"), get_named_regions(wb)$name)
+
+  expect_equal(
+    wb_to_df(wb, definedName = "iris_tab"),
+    wb_to_df(wb, definedName = "iris")
```
JanMarvin, line 427: Might want to replace definedName with namedRegion, because now it is a namedRegion of either a definedName or a table.

PR #198, opened 2022-05-23
Wrap std::string in Rcpp::String

JanMarvin:


JanMarvin:
This solves my import issues. Now I can import xlsx files with German umlauts in R 4.1 (not entirely happy with the string wrapping, but it works ...).

Another issue is file writing. This is affected in a similar way and I'm not sure how to deal with it.

PR #199, opened 2022-05-23
[WIP] Attempt to write UTF-8 from non UTF-8 systems. Did not test this

JanMarvin:
The idea is to use R's iconv in Rcpp pre writing, if the system indicated that it does not use UTF-8. Always. No matter what. For some reason `write_file()` worked fine, but the worksheet contained broken unicode:

```R
# to test
wb <- wb_workbook()$
  add_worksheet("höhö")$
  add_data("höhö", "bää", colNames = FALSE)
```


JanMarvin:
This works on my system. Therefore merging. It should not affect people on Unicode at all.

Issue #200, opened 2022-05-24
imported xlsm shoud write temp xlsm for wb$open()


JanMarvin:
[macro.xlsm.zip](https://github.com/JanMarvin/openxlsx2/files/8765196/macro.xlsm.zip)

If some xlsm file is loaded it will not open if opened with `wb$open()` due to the file ending being xlsx. If xlsm is imported, if should be written as such. Might simply check if `wb$vbaProject` contains a binary file. (Might also be beneficial, if we ever want to be able to copy bin files from one workbook to another.) Otherwise we would have to store the file ending of the imported workbook.

```R
# get and unzip the file
tmp <- tempdir()
setwd(tmp)
curl::curl_download("https://github.com/JanMarvin/openxlsx2/files/8765196/macro.xlsm.zip", "macro.xlsm.zip")
unzip("macro.xlsm.zip", exdir = ".")

# load and open it
wb <- wb_load("macro.xlsm")
wb$open()
```
<img width="575" alt="Screenshot 2022-05-24 at 20 11 39" src="https://user-images.githubusercontent.com/1645626/170103939-e3541369-2bb1-4616-beee-e642d366da83.png">

jmbarbone:
We can add that `vbaProject` check inside `$open()` and then probably have a separate `temp_xlsm()` function (same as `temp_xlsx()`) for the file path.  It might also be useful to have a check in `$save()` to give a warning when the file extension is `xlsx` but there's a bin file in `vbaProject.

Maybe a `private$macro_enabled` field would be good to simplify the checks.

JanMarvin:
I've pushed a poor mans solution in #202, feel free to improve this :)

PR #202, opened 2022-05-25
[R/load] guard against broken vml. closes #201 and #200

### LINKED ISSUE #201: vml must not be valid xml
vml might contain a div node which could contain a `<br>` node. XML parsers will bail over this. We bail over this.

JanMarvin:
skip names assignment for conditional formatting, this was broken too.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Fixes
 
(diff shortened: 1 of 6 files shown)


JanMarvin:
Example file is in #201

PR #209, opened 2022-05-28
[R/save] check that comments are saved correctly

JanMarvin:
I've crafted a new example file that contains both issues.

1. the comment counting is off. we load `comments1.xml` and `comments2.xml`. Yet we try to save them with numbers matching their sheet index
2. the comments are messed up internally


tests/testthat/test-class-comment.R
```diff
@@ -54,3 +54,19 @@ test_that("comments", {
   expect_silent(wb_save(wb, tmp))
 
 })
+
+
+test_that("load comments", {
+
+  fl <- system.file("extdata", "pivot_notes.xlsx", package = "openxlsx2")
```
JanMarvin, line 61: Wonderful, did not add it and purged my folder 😢

R/class-workbook.R
```diff
@@ -1004,16 +1004,16 @@ wbWorkbook <- R6::R6Class(
...
       ## xl/comments.xml
       if (nComments > 0 | nVML > 0) {
+        sel <- vapply(wb$comments, function(x) length(x) > 0, NA)
+        comments <- self$comments[sel]
         # TODO use seq_len() or seq_along()?
-        for (i in seq_len(nSheets)) {
-          if (length(self$comments[[i]])) {
-            fn <- sprintf("comments%s.xml", i)
+        for (i in seq_along(comments)) {
+          fn <- sprintf("comments%s.xml", i)
```
JanMarvin, line 1011: This is no longer valid. If we remove a comment 2 of three, the remaining comment will still be named 1 and 3 and worksheet_rels still has these entries. Guess we need to select worksheet_rels comment names once again.

JanMarvin:
This is still a little work in progress, I've added another comment that needs to be resolved prior merging.

@jmbarbone part of the mess is caused by me removing strict naming conventions. This was previously a bit broken in openxlsx and (hopefully) is now a bit better implemented, though it is still far from what I want. I'd prefer a reference manager counting what goes in and where and collecting everything when writing the final output. Having all these tingling relationships hardcoded in various files is so annoying. Though fixing it correctly will be another mammoth of a task. Especially because we have references to various files in different places.

Oh and once this is merged, comment class will need another cleanup. we somehow always duplicate authors entry.

JanMarvin:
Oh and once this and the encoding part are merged, I'd like to make a minor version bump. Hopefully we've ironed out a good fraction of initial issues

PR #210, opened 2022-05-28
176 options

### LINKED ISSUE #176: Review `openxlsx_options()`
We have this copied into everything.  I haven't really touched this (keeping code as is for safety) but I think we're too loose with it.

```r
openxlsx_options <- function() {
  # op <- openxlsx_options()
  # on.exit(options(op), add = TRUE)
  options(
    # increase scipen to avoid writing in scientific
    scipen = 200,
    OutDec = ".",
    digits = 22
  )
}
```

I believe these settings are to ensure that when we write numeric data that we aren't losing any accuracy.  We can probably remove it in most cases and retain this somewhere in the write functions.

jmbarbone:
Cleans up some of the options settings.  May need to take a closer look at some of the outputs.  I think the options only matter when writing data, so it should be safe enough to remove from everything else.

Resolves #176


JanMarvin:
Thanks!

PR #211, opened 2022-05-28
update codecov.yml

JanMarvin:


JanMarvin:
As requested in  https://github.com/JanMarvin/openxlsx2/pull/210
Found it here https://community.codecov.com/t/cannot-disable-codecov-patch-check/682/23

Is that what you want @jmbarbone ? They have more options over here: https://docs.codecov.com/docs/commit-status

jmbarbone:
@JanMarvin this could be what I wanted.  As long as the code coverage keeps updating we can lightly monitor it and make sure we're hitting targets.  Don't need the coverage failing because I condense some lines...

Also really great that we're at about 83% 🎉

PR #212, opened 2022-05-29
update theme

JanMarvin:
Updates the theme used in newly created workbooks. Seen some stackoverflow that we apply older table styles, when writing xlsx files. ~~This is the single use case I checked. Maybe some other color schemes are different too.~~

Old:
Screenshot 2022-05-29 at 12 15 04

Screenshot 2022-05-29 at 12 26 46

New:
Screenshot 2022-05-29 at 12 14 49

Screenshot 2022-05-29 at 12 27 00


JanMarvin:
The input is from an empty MS365 xlsx file

JanMarvin:
should be safe. we can always revert if somethings off

PR #213, opened 2022-05-29
[do_not_merge] read file with x namespace

JanMarvin:
A not for merge PR to read this file: https://github.com/ycphs/openxlsx/issues/344

Reading strange files often helps improving the reader and fixing the logic. This made me realise that our sheet ordering logic was flawed and also it is raining all day.

### DIFF
--- a/tests/testthat/test-conditional_formatting.R
+++ b/tests/testthat/test-conditional_formatting.R
@@ -249,11 +249,12 @@ test_that("conditional formatting", {
 
   expect_silent(wb1 <- wb_load(tmp))
 
-  # all.equal(wb, wb1)
-  for (sheet in seq_along(wb$sheet_names))
-    expect_equal(
-      read_xml(wb$worksheets[[sheet]]$conditionalFormatting, pointer = FALSE),
-      read_xml(wb1$worksheets[[sheet]]$conditionalFormatting, pointer = FALSE)
-    )
+  ## Test fails because of reading and writing
+  # # all.equal(wb, wb1)
+  # for (sheet in seq_along(wb$sheet_names))
+  #   expect_equal(
+  #     read_xml(wb$worksheets[[sheet]]$conditionalFormatting, pointer = FALSE),
+  #     read_xml(wb1$worksheets[[sheet]]$conditionalFormatting, pointer = FALSE)
+  #   )
 
 })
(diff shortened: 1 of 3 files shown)


tests/testthat/test-conditional_formatting.R
```diff
@@ -249,11 +249,12 @@ test_that("conditional formatting", {
...
-      read_xml(wb$worksheets[[sheet]]$conditionalFormatting, pointer = FALSE),
-      read_xml(wb1$worksheets[[sheet]]$conditionalFormatting, pointer = FALSE)
-    )
+  ## Test fails because of reading and writing
+  # # all.equal(wb, wb1)
+  # for (sheet in seq_along(wb$sheet_names))
+  #   expect_equal(
+  #     read_xml(wb$worksheets[[sheet]]$conditionalFormatting, pointer = FALSE),
+  #     read_xml(wb1$worksheets[[sheet]]$conditionalFormatting, pointer = FALSE)
+  #   )
```
JanMarvin, line 258: Assume this is due to the read/write logic. Somehow something in conditionalFormatting changed, but was not keen enough to find out. After all this is just a PR to toy around.

JanMarvin:
if my style.xml contained values, Excel was unable 5o open the file. Might be related to a missing theme?

JanMarvin:
Closing this. I don't want it merged, even though the changes are small, they are quite exotic and so far only useful in a single use case.

elinw:
It's not really that rare, it seems to happen when files are downloaded (by default they are put into a protective mode).  My example is using files from OneDrive.

JanMarvin:
Hi @elinw , you can confirm that it's the same issue, solved by this PR? And by downloaded, what do you mean downloaded from where and protected by what? The example file is constructed using some third party software. Excel itself does not do it and onedrive should not either (at least I hope so, nobody needs a cloud storage messing with it's content).

elinw:
Yes I will dig into it more in the morning, but I saw a couple of StackOverflow posts about this error message that diagnosed the x namespace as the cause  of the issue and "downloading from the internet" as the cause. I just started working with the Microsoft365R package and this happened.I'm also going to post something in their tracker. I will let you know if I find out anything.

JanMarvin:
Hm strange, thanks for pointing it out. Maybe we need to revisit that PR :smiling_face_with_tear: . It would be great if we had some more sample files to investigate the behavior. So far I was under the impression that this is only caused by this particular third party software.

I did not want to merge this for two reasons:

1. we have to check every file for the `x:` namespace before reading it, whereas we only saw it in a very specific file, and we don't even know if the namespace is always `x` or if `x` was just randomly chosen by very specific third-party software
2. we change every file without any further checks (and at least one of our current tests fails). While it might work when reading, it might fail when writing back to xlsx/xlsm.

PR #219, opened 2022-06-04
[WIP] provide charts fields

JanMarvin:
⚠️ Currently just something I was toying around with and still very early work in progress ⚠️ 

## description

Below code is a minor modification from an MS365 univariate barplot. If the code below is run, a barplot appears in the xlsx file.

## why?

It would be nice to create or modify charts from within `openxlsx2`. Unfortunately it is somewhat unlikely that anyone of us ever finds the time or motivation, to research the entire logic of the openxml chart format. Fortunately, we can create a few examples in spreadsheet software, pick up the xml files and slightly modify it. Below code does just that. It uses a barplot created on a small dataset of 5 observations. I've imported the required files from its unzipped xlsx file and recreated on our end, what is needed to let excel know where our data is located.

## but?

Yes, the but part is quite long tbh. I'm actually jumping the gun here quite a bit. First we should provide the backend for importing charts, therefore create some logic how to handle the imports instead of our current approach of simply coping the files. Afterwards the larger xml files below should be structured. At least into parts we need or can modify and other parts we cannot modify or do not understand. Right now I just had a look at some minor things. Maybe we can craft various chart types with just minor modifications of the xml.

It is not tweaked any further, therefore right now only a single barplot can be added to a worksheet.

## the code
Create a fresh workbook, add two sheets and barcharts for both sheets
```R
wb <- wb_workbook()$
  add_worksheet("S1")$add_data("S1", cars)$
  add_worksheet("S2")$add_data("S2", mtcars, rowNames = TRUE)

# create Chart input
srhc <- list(
  list(sheet = "'S2'",
       head = "$B$1",
       body = "$B$2:$B$33",
       cat = "$A$2:$A$33")
  ,list(sheet = "'S2'",
        head = "$C$1",
        body = "$C$2:$C33",
        cat = "$A$2:$A$33")
)

wb <- wb_add_chart(wb, sheet = 2, srhc, dims = "A1:H18")



srhc <- list(
  list(sheet = "'S1'",
       head = "$B$1",
       body = "$B$2:$B$51",
       cat = "")
)

wb <- wb_add_chart(wb, sheet = 1, srhc, dims = "A1:H18")

wb$open()
```

## Screenshots
Screenshot 2022-10-02 at 13 57 55
Screenshot 2022-10-02 at 13 58 17


JanMarvin:
first update:
* minor cleanup of xml
* create barplot of multiple variables

JanMarvin:
second update 
* add category variable to add to openxml bargraph. this is the x axis in the plot. example plots two variables from mtcars using rownames as category

JanMarvin:
Third update:
* reading charts during import is now supported
* a dummy function to create a barplot is provided

JanMarvin:
I plan to move the add_chart part into another pull request and port the chart fields part into main. It should not do any harm (might be a fraction of a millisecond slower), but we benefit from this.

1. we are actually importing the chart file.
2. we can avoid some duplicate entries in Content_Types and other rels.

JanMarvin:
Fourth update:
* rebase on main
* update examples
* add screenshots

PR #220, opened 2022-06-05
[R/add_border] add_border and wb_add_border

JanMarvin:
Create borders for cell regions. This creates border styles for a specific region.

* [x] add tests
* [x] merge styles if possible. If cell as `<xf../>` node, copy the node and applyBorder and borderid. Otherwise if two or more cells share the style, they all would get the border.
* [x] inner grid (maybe providing a single style)


JanMarvin:
unrelated fix for showGridLines option and newly introduced rowColHeaders option.

JanMarvin:
Any clue why the test are complaining @jmbarbone ?

jmbarbone:
> Any clue why the test are complaining @jmbarbone ?

@JanMarvin, it looks like some weird `roxygen2` behavior where the the lines changes in 
237a1bd were being appended to the documentation.

JanMarvin:
Thanks, didn't see that. Maybe because of the example in the `wbWorkbook` part

PR #227, opened 2022-06-09
214 conditional formatting

### LINKED ISSUE #214: Clean up `wb_conditional_formatting()`
This isn't a proper wrapper.  Must have missed it because it's saved in a separate file.

Related to #75

jmbarbone:
resolves #214


R/class-workbook.R
```diff
@@ -5516,36 +5442,33 @@ wbWorkbook <- R6::R6Class(
...
         ),
 
-        ## topN ----
         topN = sprintf(
           '<cfRule type="top10" dxfId="%s" priority="1" rank="%s" percent="%s"></cfRule>',
           dxfId,
-          values[1],
-          values[2]
+          values$rank,
+          if (values$percent == 1) 1 else "NULL"
```
JanMarvin, line 5683: not sure if null is a valid option here. would have expected `0`

R/utils.R
```diff
@@ -89,6 +89,15 @@ dir_create <- function(..., warn = TRUE, recurse = TRUE) {
   path
 }
 
+as_binary <- function(x) {
```
JanMarvin, line 92: suggestion: `as_bool()`

tests/testthat/test-conditional_formatting.R
```diff
@@ -1,259 +1,394 @@
...
+    rows = 2:11,
+    style = posStyle,
+    type = "topN",
+    params = list(rank = 20, percent = TRUE)
+  )
 
   exp <- c(
-    `A2:A11` = "<cfRule type=\"top10\" dxfId=\"1\" priority=\"2\" rank=\"5\" percent=\"NULL\"></cfRule>",
-    `B2:B11` = "<cfRule type=\"top10\" dxfId=\"1\" priority=\"1\" rank=\"20\" percent=\"1\"></cfRule>"
+    `A2:A11` = '<cfRule type="top10" dxfId="1" priority="2" rank="5" percent="NULL"></cfRule>',
```
JanMarvin, line 337: oh it is `NULL` here

tests/testthat/test-conditional_formatting.R
```diff
@@ -1,259 +1,394 @@
...
-  posStyle <- create_dxfs_style(font_color = c(rgb = "FF006100"), bgFill = c(rgb = "FFC6EFCE"))
+expect_save <- function(wb) {
+  # Quick function to check original saving.
+  assert_workbook(wb)
+  path <- temp_xlsx()
 
-  wb$styles_mgr$styles$dxfs <- c(wb$styles_mgr$styles$dxfs,
-                                 c(negStyle, posStyle)
-  )
+  expect_silent(wb_save(wb, path))
```
JanMarvin, line 18: maybe needs `testthat::`

R/class-workbook.R
jmbarbone: `0` seems better.  Our tests were expecting `"NULL"` but maybe that wasn't the best.

R/utils.R
jmbarbone: Well, it does convert it to either `0` or `1`, rather than `TRUE` or `FALSE`?

tests/testthat/test-conditional_formatting.R
jmbarbone: Works fine with `"0"` and is easier to handle

tests/testthat/test-conditional_formatting.R
jmbarbone: Eh, anything within `tests/testthat/` can probably safely assume `testthat` is loaded.

JanMarvin: dunno, but it's a function using testthat wrappers and lintr complains. Therefore my idea, but we might as well: not care ;)

R/utils.R
JanMarvin: ha, yeah, in c it would be a bool, but it's fine the way it is.

tests/testthat/test-conditional_formatting.R
jmbarbone: Fair enough.  I was just trying to be a little lazy.

R/class-workbook.R
```diff
@@ -5400,6 +5728,281 @@ wbWorkbook <- R6::R6Class(
...
+          ## formula contains the colours
+          ## values contains numerics or is NULL
+          ## dxfId is ignored
+
+          if (is.null(values)) {
+            # could use a switch() here for length to also check against other
+            # lengths, if these aren't checked somewhere already?
+            if (length(formula) == 2L) {
+              sprintf(
+                # TODO is this indentation necessary?
```
JanMarvin, line 5789: Haha, was starring at it and thought the same. No, absolutely not. Can be indented in a sane way, or not at all.

R/class-workbook.R
```diff
@@ -5400,6 +5728,281 @@ wbWorkbook <- R6::R6Class(
...
+                            </dataBar>
+                            <extLst><ext uri="{B025F937-C7B1-47D3-B67F-A62EFF666E3E}" xmlns:x14="http://schemas.microsoft.com/office/spreadsheetml/2009/9/main">
+                            <x14:id>{%s}</x14:id></ext></extLst></cfRule>',
+              params$showValue,
+              values[[1]],
+              values[[2]],
+              posColour,
+              guid
+            )
+          }
```
JanMarvin, line 5905: If I'm not mistaken, this could be shortened. If the value part `<cfvo type="num" val="%s"/><cfvo type="num" val="%s"/>` is the only difference, we could create it only if `!is.null(values)` and sprintf this string or otherwise just `""`. But I haven't touched it previously, we might as well ignore it for the future.

R/class-workbook.R
```diff
@@ -5400,6 +5728,281 @@ wbWorkbook <- R6::R6Class(
...
+          sprintf(
+            '<cfRule type="containsText" dxfId="%s" priority="1" operator="containsText" text="%s">
+                        	<formula>NOT(ISERROR(SEARCH("%s", %s)))</formula>
+                       </cfRule>',
+            dxfId,
+            values,
+            values,
+            # is this unlist correct?  Would this not work?
+            # > strsplit(sqref, split = ":")[[1]]
+            unlist(strsplit(sqref, split = ":"))[[1]]
```
JanMarvin, line 5933: Agree, looks a bit strange. I felt that this entire code section was a bit fragile when I went over it a little while ago. Minor changes have a tendency to cause broken xlsx files.
This will actively select only the first element. It might be that the input is `A:A` and we only want `A`)

R/class-workbook.R
```diff
@@ -5400,6 +5728,281 @@ wbWorkbook <- R6::R6Class(
...
+
+        bottomN = {
+          sprintf(
+            '<cfRule type="top10" dxfId="%s" priority="1" rank="%s" percent="%s" bottom="1"></cfRule>',
+            dxfId,
+            values$rank,
+            values$percent
+          )
+        },
+        # do we have a match.arg() anywhere or will it just be showned in this switch()?
```
JanMarvin, line 5997: Either it is shown here or I the stop condition might never be reached. Unsure

R/helperFunctions.R
```diff
@@ -128,8 +128,18 @@ getId <- function(x) reg_match0(x, '(?<= Id=")[0-9A-Za-z]+')
 #' @keywords internal
 #' @noRd
 validateColour <- function(colour, errorMsg = "Invalid colour!") {
+  colour <- check_valid_colour(colour)
```
JanMarvin, line 131: Oh I always wanted to get rid of the `colours` in favor of `colors`

R/helperFunctions.R
```diff
@@ -128,8 +128,18 @@ getId <- function(x) reg_match0(x, '(?<= Id=")[0-9A-Za-z]+')
...
+  if (isFALSE(colour)) {
+    stop(errorMsg)
+  }
+
+  colour
+}
+
+check_valid_colour <- function(colour) {
+  # Not proud of this.  Returns FALSE if not a vaild, otherwise  cleans up.
+  # Probably not the best, but working within the functions we alreayd have.
```
JanMarvin, line 142: We can always improve in the future :)

vignettes/conditional-formatting.Rmd
```diff
@@ -0,0 +1,342 @@
...
+  collapse = TRUE,
+  comment = "#>"
+)
+set.seed(123)
+```
+
+```{r setup}
+library(openxlsx2)
+```
+
```
JanMarvin, line 21: We can enrich this with a few screenshots sometime in the future.

JanMarvin (approved):
Looking quite well already!

R/class-workbook.R
jmbarbone: Okay, that's good to know.  I couldn't tell if the additional spacing was necessary or just from some copy-paste.  We can probably safely clean this up later.

JanMarvin: Sure!

R/class-workbook.R
jmbarbone: Ah, I meant `strsplit(sqref, split = ":")[[1]][1]`, which makes a bit more sense.

R/class-workbook.R
jmbarbone: I had added a `match.arg()` in `$add_conditional_formatting()` so we can change this to an internal error in case something really goes wrong

R/helperFunctions.R
jmbarbone: I've very prone to using `colours`.  Might be worth it to go through all the code/documentation and use just `colors` for everything.

JanMarvin:
What's the status of this @jmbarbone ? Can it be merged?

jmbarbone:
I think it's good to merge now.  Don't think there's anything pending.

JanMarvin:
Thanks! Merged.

PR #233, opened 2022-06-14
cloneSheetStyle -> wb_clone_sheet_style

JanMarvin:
Was trying to add a quick test for `cloneSheetStyle()`, but that did not work as expected.

* [x] this is not yet a wrapper
* [x] if not all cloned style cells are available on the new sheet, we should add them
* [x] create tests


JanMarvin:
this needs a news entry, otherwise it's good to go

PR #236, opened 2022-06-18
[R/write] add dims argument to write

JanMarvin:
Initial implementation for `dims` in write. In the future this could be improved to check if the data we attempt to add to the worksheet matches the `dims`. This is somewhat required to add data to a named region


R/class-workbook.R
```diff
@@ -805,6 +806,7 @@ wbWorkbook <- R6::R6Class(
     add_data = function(
         sheet           = current_sheet(),
         x,
+        dims            = NULL,
         startCol        = 1,
```
jmbarbone, line 810: `dims` after `startCol`, `startRow` seems better when using unnamed arguments:

```r
wb$add_data(sheet, data, 1, 1) # col, row
wb$add_data(sheet, data, range = c(1, 1) # explicitly use range
```

R/utils.R
```diff
@@ -108,3 +108,31 @@ random_string <- function(size = NULL) {
 
   substr(res, 1L, size)
 }
+
+dims_to_rowcol <- function(x, as_integer = FALSE) {
```
jmbarbone, line 112: Maybe think about the reverse: `row_col_to_dim()`?  This would allow us to do:

```r
#' @param row,col,dims Start col and rows for data.  When `dims` is specified,
#'   `row` and `col` are ignored.
foo <- function(wb, row = NULL, col = NULL, dims = row_col_to_dim(row, col)) {
  # pass row and col directly to dims, prioritize over the individual params
  dims <- validate_dims(dims)
  dims$row
  dims$col
  ...
}
```

R/class-workbook.R
```diff
@@ -854,6 +858,7 @@ wbWorkbook <- R6::R6Class(
     add_data_table = function(
         sheet       = current_sheet(),
         x,
+        dims        = NULL,
         startCol    = 1,
         startRow    = 1,
         xy          = NULL,
```
jmbarbone, line 864: What are we doing with `xy`?  Is this doing to be deprecated in favor of `dims`?

jmbarbone (commented):
We have an `xy` param (the name I don't like very much) which might need to be deprecated in favor of `dims`.  I think we have some potential to use `dims` (or maybe a `range`)  param in other functions.  We have some helpers for converting these so many we can make a special S3 `openxlsx2_cell_range` class that allows these things to be saved/passed more efficiently.  Maybe look at [`cellranger`](https://github.com/rsheets/cellranger) for some inspiration (if not dependency).

```r
str(cellranger::cell_limits(ul = c(1, 2), lr = c(7, 6)))
#> List of 3
#>  $ ul   : int [1:2] 1 2
#>  $ lr   : int [1:2] 7 6
#>  $ sheet: chr NA
#>  - attr(*, "class")= chr [1:2] "cell_limits" "list"
```

R/class-workbook.R
JanMarvin: good thinking. I never had a use for `xy`. Guess we should deprecate it.

R/class-workbook.R
JanMarvin: yeah, I was looking for a good spot, after startCol, startRow is a good idea too. I'll switch it

R/utils.R
JanMarvin: indeed! nice, I like it. will look at it

JanMarvin:
> We have some helpers for converting these so maybe we can make a special S3 `openxlsx2_cell_range` class that allows these things to be saved/passed more efficiently. Maybe look at [`cellranger`](https://github.com/rsheets/cellranger) for some inspiration (if not dependency).
> 
> ```r
> str(cellranger::cell_limits(ul = c(1, 2), lr = c(7, 6)))
> #> List of 3
> #>  $ ul   : int [1:2] 1 2
> #>  $ lr   : int [1:2] 7 6
> #>  $ sheet: chr NA
> #>  - attr(*, "class")= chr [1:2] "cell_limits" "list"
> ```

Ha, that sounds like over-engineering 😄
I'd like to have it a bit better integrated, but I doubt that it requires another package for that. That `range` would be a better name than `dims` ... let's say I realized that a bit to late.

R/utils.R
jmbarbone: The `validate_dims()` here is where I'd imagine we could implement some of those `cellranger`-like things.  Then we have a good helper function to consistently use for any instances in which we need to establish cell ranges and not have to worry about passing integers or `"A1:C3"` specs or anything.

JanMarvin: The character/integer part was solely for our output. In `update_cell()` I allow to use dims to create data on a worksheet. Iirc I use rows and cols to identify areas that are available and must be created or can simply be modified. Though, in an ideal world, I would want to ditch most of `update_cell()`s code and use what is available in `wide_to_long()` and a do a full join afterwards. Updating the existing cells and adding the newly created. Though that will probably be a task for a rainy day in the future. I really dislike the duplicated code there. It's among the earliest newly created code for `openxlsx2` and... not the best :)

JanMarvin: Initial draft is added. Calling it a day

tests/testthat/test-utils.R
```diff
@@ -13,3 +13,36 @@ test_that("rbindlist", {
...
+  exp <- list(1:2, 1:2)
+  got <- dims_to_rowcol("A1:B2", as_integer = TRUE)
+  expect_equal(exp, got)
+
+  exp <- list(1:2, c(1L))
+  got <- dims_to_rowcol("A:B", as_integer = TRUE)
+  expect_equal(exp, got)
+
+  exp <- list("A", c("1"))
+  got <- dims_to_rowcol("A:A", as_integer = FALSE)
```
JanMarvin, line 33: If we ever use this to create the rows to write to, rows should be something else. Maybe return `Inf` or something. In Excel logic this will use as many rows as are available.

R/write.R
```diff
@@ -589,6 +580,7 @@ write_data_table <- function(
     x,
     startCol = 1,
     startRow = 1,
+    dims,
```
JanMarvin, line 583: dims = NULL,

JanMarvin:
I intend to merge as it is, ofc fixing the other remaining dim in write, but won't follow through further improvements in regards of Cellranger mimicking. It's not that I don't want the feature, it's just that I want dims more and have other things as well. Hmpf

PR #240, opened 2022-06-20
remove `wb_get_sheet_name`

JanMarvin:
There is no real need for this function


R/class-workbook.R
```diff
@@ -3613,7 +3613,7 @@ wbWorkbook <- R6::R6Class(
           ref1 = paste0("$", cols[1]),
           ref2 = paste0("$", cols[2]),
           name = "_xlnm.Print_Titles",
-          sheet = names(self)[[sheet]],
+          sheet = self$get_sheet_names()[[sheet]],
```
JanMarvin, line 3616: Wild guess, but maybe RGui was picking this up?

R/class-workbook.R
```diff
@@ -5612,35 +5612,6 @@ lcr <- function(var) {
...
-# TODO Does this need to be checked?  No sheet name can be NA right?
-# res <- self$sheet_names[ind]; stopifnot(!anyNA(ind))
-
-#' Get sheet name
-#'
-#' @param wb a [wbWorkbook] object
-#' @param index Sheet name index
-#' @return The sheet index
-#' @export
-wb_get_sheet_name = function(wb, index = NULL) {
```
JanMarvin, line 5625: Do we need a deprecated for this, @jmbarbone ?

R/class-workbook.R
JanMarvin: If this does not fix it, we need it .... grml

R/class-workbook.R
JanMarvin: I gave up. No need to clean this up right now

PR #247, opened 2022-06-23
re-enable oldrel-1

jmbarbone:
Well, `R 4.2.1` was released today so maybe our `oldrel-1` will be valid again.

Couldn't figure out what was causing this to fail in `R 4.1.3`

### DIFF
--- a/.github/workflows/check-standard.yaml
+++ b/.github/workflows/check-standard.yaml
@@ -20,7 +20,7 @@ jobs:
         config:
           - {os: macOS-latest,   r: 'release'}
           - {os: windows-latest, r: 'release'}
-          # - {os: windows-latest, r: 'oldrel-1'} # i386 multiarch test fails
+          - {os: windows-latest, r: 'oldrel-1'}
           - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
           - {os: ubuntu-latest,   r: 'release'}
           - {os: ubuntu-latest,   r: 'oldrel-1'}
--- a/openxlsx2.Rproj
+++ b/openxlsx2.Rproj
@@ -17,6 +17,7 @@ StripTrailingWhitespace: Yes
 
 BuildType: Package
 PackageUseDevtools: Yes
+PackageCleanBeforeInstall: Yes
 PackageInstallArgs: --no-multiarch --with-keep.source
 PackageCheckArgs: --as-cran
 PackageRoxygenize: rd,collate,namespace,vignette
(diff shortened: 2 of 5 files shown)


JanMarvin:
Might as well be a bug in testthat or any other of the packages it uses, looking at the testthat bug tracker there are Unicode issues concerning brio.

It's a bit unfortunate that the tests fail, because I wanted to  especially test non unicode encodings. Simply assuming that it works is a bit negligent. Even though R 4.2 is now oldrel, I assume that R < 4.2 will be around for quite a long time. Some people still run R 3.6 which is quite dated. And aside from the English speaking world, non ASCII characters are quite common 😬

JanMarvin:
There are at least two errors

## the failing tests
```txt
-- Error (Line 8): read html source without r attribute on cell ----------------
<std::bad_alloc/C++Error/error/condition>
Error in `loadvals(wb$worksheets[[i]]$sheet_data, worksheet_xml)`: std::bad_alloc
Backtrace:
 1. openxlsx2::wb_load("https://github.com/JanMarvin/openxlsx2/files/8702731/fichier_complementaire_ccam_descriptive_a_usage_pmsi_2021_v2.xlsx")
 2. openxlsx2::loadvals(wb$worksheets[[i]]$sheet_data, worksheet_xml)
      at openxlsx2/R/wb_load.R:624:6

Error in reporter$stop_if_needed() : Test failed
```

The second utils-helper test. Which is equally strange.

## the ugly hotfix

I can trigger the tests to fail, once I run `devtools::load_all(":")`. Maybe some strange interaction with `testthat` or with base R?
We could simply skip over the tests, but I'm not yet seeing, where or why we fail.

```R
  if (.Platform$r_arch == "i386" && .Platform$OS.type == "windows")
    skip("Skip test on Windows i386")
```

JanMarvin:
well, check ran fine as assumed. Though no clue what is going on. Maybe `i386` and `x86_64` somehow behave differently? Especially the `rbindlist(data.frame())` crash is strange and it might as well be a related issue, even though it points to `loadvals()`.

JanMarvin:
Who would have thought it! It was a bug, at least it was a bug in R 4.1 i386 and it is fixed now. And it only took some 6-8 hours. No clue why the pkgdown build is failing, but Windows is now fine.

JanMarvin:
I high jacked your pull request, but maybe you can have a second look and squash merge it if you are fine

jmbarbone:
I guess the pkgdown issue will be resolved with #248.  Other than that it seems fine.

PR #248, opened 2022-06-25
[manual] check for ggplot2

JanMarvin:
`pkgdown` bails if `ggplot2` is not found. it is used only in this example file of a single package, therefore I do not want to add it as a dependency. We could add it to the `pkgdown` github action, but after all the claim is valid.

### DIFF
--- a/R/class-workbook-wrappers.R
+++ b/R/class-workbook-wrappers.R
@@ -707,6 +707,7 @@ wb_remove_row_heights <- function(wb, sheet = current_sheet(), rows) {
 #' wb$add_worksheet("Sheet 1", gridLines = FALSE)
 #'
 #' ## create plot objects
+#' if (find.packages("ggplot2", quiet = TRUE)) {
 #' require(ggplot2)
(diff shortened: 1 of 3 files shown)


R/class-workbook-wrappers.R
```diff
@@ -707,6 +707,7 @@ wb_remove_row_heights <- function(wb, sheet = current_sheet(), rows) {
 #' wb$add_worksheet("Sheet 1", gridLines = FALSE)
 #'
 #' ## create plot objects
+#' if (find.packages("ggplot2", quiet = TRUE)) {
```
jmbarbone, line 710: Change to `if (requireNamespace("ggplot2", quietly = TRUE))`?  This is what CRAN might expect.

JanMarvin: Thanks, I'll try it tomorrow

JanMarvin:
The upstream bug report: https://github.com/r-lib/pkgdown/issues/2145

PR #253, opened 2022-06-27
release 0.2.1

JanMarvin:
Bump version 🚀 

There have been many, many, many improvements over our first public release and it is time that our release reflected that.


jmbarbone:
Do we want to just call this `0.2.1` and lead up to `0.3.0` for a CRAN release?

JanMarvin:
We can do so

PR #255, opened 2022-06-28
[R/write] fixes for NA, NaN, and Inf. Closes #254

### LINKED ISSUE #254: capture infinite and nan when writing
Similar to `NA`, we might end up with `Inf` or `NaN` when writing numeric data. These values should be treated as exceptions `#VALUE!` and/or `#NUM!` or as characters.

JanMarvin:
This fixes the read/write of NA, NaN and Inf. I'm not sure yet if it's a good idea to always convert `#NUM!` and/or `#VALUE!` and maybe more openxml exceptions to R's NaN, NA or Inf. All these exceptions have special meaning in R and must not match their openxml counterparts. I am open to arguments as to why we should or should not do this.

### DIFF
--- a/R/write.R
+++ b/R/write.R
@@ -167,8 +167,13 @@ update_cell <- function(x, wb, sheet, cell, data_class,
 
         # for now convert all R-characters to inlineStr (e.g. names() of a data frame)
         if ((data_class[m] == openxlsx2_celltype[["character"]]) || ((colNames == TRUE) && (n == 1))) {
-          cc[sel, "c_t"] <- "inlineStr"
-          cc[sel, "is"]   <- paste0("<is><t>", as.character(value), "</t></is>")
+          if (is.na(value)) {
+            cc[sel, "v"] <- "#N/A"
+            cc[sel, "c_t"] <- "e"
+          } else {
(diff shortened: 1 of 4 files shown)


R/write.R
```diff
@@ -503,7 +514,7 @@ write_data2 <-function(wb, sheet, data, name = NULL,
     }
 
     sel <- which(dc == openxlsx2_celltype[["character"]]) # character
-    for (i in sel) {
+    if (length(sel)) {
```
JanMarvin, line 517: this looks pretty stupid to be honest 🤦

tests/testthat/test-read_from_created_wb.R
```diff
@@ -40,39 +40,47 @@ test_that("Reading NAs and NaN values", {
   fileName <- file.path(tempdir(), "NaN.xlsx")
   na.string <- "*"
 
+  wb <- wb_workbook()
+
   ## data
-  a <- data.frame(
+  A <- data.frame(
```
JanMarvin, line 46: I assume this test was copied over from `openxlsx` (at least I hope nobody here uses `c` as object name) and shortened when we did not support all features. Now that we support more features, we might want to add the missing pieces.

PR #268, opened 2022-07-10
[R/cf] cleanups and outsourcing

JanMarvin:
* removed old code
* provide a new function for `read_xml()` to read or trim whitespaces (read is the default. in worksheets we sometimes end up with `<is><t> </t></is>`.) Initially I wanted to trim whitespaces on the R level, but that did not work out at fist try
* outsourced code from the huge switch condition into smaller functions that we could test easily. maybe we could reduce these into a general function, but I was worried that I would spend the rest of my Sunday on this (hint: I did, but for other reasons).


JanMarvin:
@jmbarbone increased the coverage a bit, there was a lot of old unused code. I didn't even notice this at first and merely wondered why my changes had no effect.

While working on this PR I snuck in a few Pugi changes, but nothing serious. When checking if my changes to handle multiple databars were still working, I spotted something else. Should our wrappers ever modify the initial workbook? I was under the impression that they should not and should be cloned if requested. Might open an issue or discussion.

JanMarvin:
Oh, my changes still work nicely. Updated code:

```R
library(openxlsx2)

wb <- wb_workbook()
wb$add_worksheet("databar")
## Databars
wb$add_data("databar", -5:5, startCol = 1)
wb_add_conditional_formatting(
  wb,
  "databar",
  cols = 1,
  rows = 1:11,
  type = "dataBar"
) ## Default colours

wb$add_data("databar", -5:5, startCol = 3)
wb_add_conditional_formatting(
  wb,
  "databar",
  cols = 3,
  rows = 1:11,
  type = "dataBar",
  params = list(
    showValue = FALSE,
    gradient = FALSE
  )
) ## Default colours

wb$add_data("databar", -5:5, startCol = 5)
wb_add_conditional_formatting(
  wb, 
  "databar",
  cols = 5,
  rows = 1:11,
  type = "dataBar",
  style = c("#a6a6a6"),
  params = list(showValue = FALSE)
)

wb$add_data("databar", -5:5, startCol = 7)
wb_add_conditional_formatting(
  wb,
  "databar",
  cols = 7,
  rows = 1:11,
  type = "dataBar",
  style = c("red"),
  params = list(
    showValue = TRUE,
    gradient = FALSE
  )
)

wb$add_data("databar", -5:5, startCol = 9)
wb_add_conditional_formatting(
  wb,
  "databar",
  cols = 9,
  rows = 1:11,
  type = "dataBar",
  style = c("#a6a6a6", "#a6a6a6"),
  params = list(showValue = TRUE, gradient = FALSE)
)

wb$open()
```

JanMarvin:
might need some style mgr functions to make this work correctly.

jmbarbone:
> Should our wrappers ever modify the initial workbook? I was under the impression that they should not and should be cloned if requested. Might open an issue or discussion.

Correct, the idea of the piped wrappers is to not modify the object in line, but require an assignment.  Did we miss a `$clone()` somewhere?

JanMarvin:
Well my code snippet above did succeed, that made me suspicious. Thanks for confirming.

JanMarvin:
This started as a smaller commit to increase the coverage of conditional formatting. I touched on a few other things along the way, but somehow most of the lines added are documentation updates and tests.

* Styles for conditional formatting now use `style_mgr` and using one requires adding `"stylename"`, whereas before the style object was pushed down. Previously it was a bit Russian roulette which style was picked up. We simply called `unique()` on a vector that could change its order and used the index position in already in our output.
* `read_xml()` slightly changed. We now use better whitespace trimming.
* Oh, and the coverage has increased!

PR #278, opened 2022-07-17
remove `$hyperlink` code

JanMarvin:
I do not see a use case for this and could not make it work, creating links from `openxlsx2`.
* Is it defunct at the moment?
* Was it never used or was it used in `openxlsx`?
* Do we need it?

### DIFF
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -5139,7 +5139,6 @@ wbWorkbook <- R6::R6Class(
         } else {
           ## Write worksheets
           ws <- self$worksheets[[i]]
-          hasHL <- length(ws$hyperlinks) > 0
 
           prior <- ws$get_prior_sheet_data()
           post <- ws$get_post_sheet_data()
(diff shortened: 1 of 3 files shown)


JanMarvin:
OK, an example is in our `loadExample.xlsx`. Have to revisit that. With `main` wbHyperlink objects are created. With this branch only xml is returned, but the links no longer work.
```XML
<hyperlink ref="B25" r:id="rId1" />
<hyperlink ref="B26" r:id="rId2" />
<hyperlink ref="B28" r:id="rId3" location="Decimal_fractions" />
<hyperlink ref="B29" r:id="rId4" location="Decimal_fractions" />
<hyperlink ref="B30" r:id="rId5" location="Decimal_fractions" />
<hyperlink ref="B33" location="IrisSample!A1" display="IrisSample!A1" />
<hyperlink ref="B34" location="IrisSample!A2" display="IrisSample!A2" />
```

PR #280, opened 2022-07-20
implement `add_sparklines()`

JanMarvin:
* [x] add wrapper
* [x] add tests
* [x] improve `extLst` handling, though that might require another PR
* [ ] add it to a vignette (it's a neat gimmick after all)


JanMarvin:
finall ... 🤦

PR #281, opened 2022-07-21
improve random strings

JanMarvin:
This is a bit tricky and I'd be open to suggestions how to improve this or if we are fine with this. I needed it, to run a test and it works well. Therefore ... another suggestion must be better :)

This approach does not alter the seed when calling `random_string()`, even with strings of random sizes. I applied these in a row, but ran into a problem, where I actually wanted the seed to be altered when calling it in `openxlsx2`, therefore I've added a way for `openxlsx2` to handle such cases.

Once `random_string()`is called:

1. We backup the current seed and check for `.openxlsx2.seed`, this is the seed we want to use. 
2. If none is found, we create one ourself.
3. Afterwards we use `stringi` to create a random string for us. 
4. We store the altered string as `.openxlsx2.seed` therefore once we call `random_string()` again, it will pick up this seed.
5. We restore the previous global seed as if it was never altered.


JanMarvin:
The onus is, we pollute the .GlobalEnv with our seed and if we want to set the seed for `random_seed()` we have to remove `.openxlsx2.seed` previously.

JanMarvin:
I use this in #280

tests/testthat/test-utils.R
```diff
@@ -23,3 +23,19 @@ test_that("create_char_dataframe", {
...
+  suppressWarnings(rm(.openxlsx2.seed, envir = .GlobalEnv))
+
+  x <- .Random.seed
+  tmp <- openxlsx2:::random_string()
+  y <- .Random.seed
+  expect_identical(x, y)
+  expect_equal("HmPsw2WtYSxSgZ6t", tmp)
+
+  x <- .Random.seed
+  tmp <- openxlsx2:::random_string(length = 6)
```
JanMarvin, line 38: it is not helpfull to switch between branches back and forth, I have removed `openxlsx2:::` already

R/utils.R
```diff
@@ -98,22 +98,40 @@ as_binary <- function(x) {
...
+#' @param length length
+#' @param pattern pattern
+#' @keywords internal
+#' @noRd
+random_string <- function(n = 1, length, pattern = "[A-Za-z0-9]") {
   # https://github.com/ycphs/openxlsx/issues/183
   # https://github.com/ycphs/openxlsx/pull/224
-  res <- basename(tempfile(""))
+  
+  if (missing(length)) length <- 16
```
jmbarbone, line 114: Just set the default value to `16`?

R/utils.R
```diff
@@ -98,22 +98,40 @@ as_binary <- function(x) {
...
 
-  if (is.null(size)) {
-    return(res)
-  }
+  rng <- get(".Random.seed", .GlobalEnv)
+
+  # try to get a previous openxlsx2 seed and use this as random seed
+  try(openxlsx2_seed <- get(".openxlsx2.seed", .GlobalEnv), TRUE)
```
jmbarbone, line 119: `get0()` is safer

jmbarbone: But we could also store it in `options()` instead -- less intrusive

R/utils.R
```diff
@@ -98,22 +98,40 @@ as_binary <- function(x) {
...
+#' random string function that does not alter the seed.
+#' 
+#' simple wrapper around `stringi::stri_rand_strings()``
+#' 
+#' @param n n
+#' @param length length
+#' @param pattern pattern
+#' @keywords internal
+#' @noRd
+random_string <- function(n = 1, length, pattern = "[A-Za-z0-9]") {
```
jmbarbone, line 110: Add param `keep_seed = FALSE` to skip over all the seed checking/changing?

R/utils.R
```diff
@@ -98,22 +98,40 @@ as_binary <- function(x) {
...
+  # try to get a previous openxlsx2 seed and use this as random seed
+  try(openxlsx2_seed <- get(".openxlsx2.seed", .GlobalEnv), TRUE)
 
-  size <- as.integer(size)
-  stopifnot(size >= 0L)
-  while (nchar(res) < size) {
-    res <- paste0(res, random_string())
+  if (exists("openxlsx2_seed")) {
+    # found one, change the global seed for stri_rand_strings
+    assign(".Random.seed", openxlsx2_seed, .GlobalEnv)
```
jmbarbone, line 123: `globalenv()` may be better to use

R/utils.R
```diff
@@ -98,22 +98,40 @@ as_binary <- function(x) {
...
 }
 
-random_string <- function(size = NULL) {
-  # creates a random string using tempfile() which does better to not affect the
-  # random seed
+#' random string function that does not alter the seed.
+#' 
+#' simple wrapper around `stringi::stri_rand_strings()``
+#' 
+#' @param n n
```
jmbarbone, line 105: `@inheritParams stringi::stri_rand_strings`

jmbarbone (changes requested):
`withr::with_seed()` also checks seed type, which we may want to do as well: https://github.com/r-lib/withr/blob/main/R/seed.R

R/utils.R
JanMarvin: eh, haha, yes. I kinda assumed that length was not required for stringi, fixed it but to tired :)

R/utils.R
JanMarvin: I had this initially, but I wasn't really happy with that. after all we are either polluting `.GlobalEnv` or `options()` and since `.Random.seed` already lives there, it felt like a better place

R/utils.R
JanMarvin: Hm, yes, can do that

R/utils.R
JanMarvin: Usually I would have used this, but when setting the seed I found this. Any specific reason?

R/utils.R
JanMarvin: Thanks, I knew there was a better way

R/utils.R
JanMarvin: This is a bit over engineering. Currently it is only used in the style helpers and in the sparkline PR. We could avoid tinkering with saving our seed in the `create_sparkline()` function if we agree that calling that function is not a harmful interaction with the seed. The purpose of this whole PR was to have a function that keeps the seed intact, but still generates various random strings if called multiple times by us, but still keeps the seed intact. Something I don't even think is necessary in the first place 🙈.

JanMarvin:
fixed the tests

JanMarvin:
We could add a seed counter option. We set and increase in random_string() and use set.seed() with it. That would be the slimmest approach, but personally I'm fine with both. The current in this PR and the sketched idea as well. I see if I have time to add this later today, otherwise I'll merge with a to-do note. It's a minor function without much use at the moment. Unless someone wants to add styles or sparklines it isn't triggered at all.

JanMarvin:
Scratch that last comment. Check complains therefore I'll change it.

```
Found the following assignments to the global environment:
File ‘openxlsx2/R/utils.R’:
  assign(".openxlsx2.seed", .Random.seed, globalenv())
```

PR #290, opened 2022-07-31
Dxf nofont

JanMarvin:
Allows this:

```R
yel_style <- create_dxfs_style(font_name = NULL, font_size = NULL)
```


JanMarvin:
To be merged after #288

PR #292, opened 2022-08-01
wbColor

JanMarvin:
Provide `wbColor`. At the moment this requires a bit more testing. The main feature is, a general color class. This improves handling indexed, theme and rgb colors as well as validated colors.

```R
wb <- wb_workbook() %>% 
  wb_add_worksheet("S1") %>% 
  wb_add_data("S1", mtcars) %>% 
  wb_add_fill(dims = "D5:J23", color = wb_color("blue")) %>% 
  wb_add_font(dims = "D5:J23", color = wb_color("gray")) %>% 
  wb_add_fill(dims = "E5:E23", color = wb_color("gray")) %>% 
  wb_add_fill(dims = "F5:F23", color = wb_color("gray", tint = 0.6)) %>% 
  wb_add_fill(dims = "G5:G23", color = wb_color(theme = "6")) %>% 
  wb_add_fill(dims = "H5:H23", color = wb_color(theme = "11")) %>% 
  wb_add_fill(dims = "I5:I23", color = wb_color(indexed = "4")) %>% 
  wb_open()
```


JanMarvin:
Maybe you can have a look @jmbarbone ? If you're short on time, a hint if you like it or not will do. This is a breaking change that introduces `wb_color()`, to simplify handling of colors.

If we want to add this, we might need a few more `if (!is.null(color)) assert_color(color)` to be on the safe side. And it's a bit inconsistent, where we apply colors and where not, but that might be possible to iron out in a follow up PR.

jmbarbone:
@JanMarvin I'll be able to take a peak this week.

Quick thoughts: Is `wbColor` needed as an `{R6}` object?  Looks mostly like a named `list` with a single xml related method.  Would there be some plans to extend the class any more?

JanMarvin:
Not really, I just to inspiration from the wb_hyperlink class, which does not really do anything either :)
If some simpler S3 function works better for you, I can switch over to that as well.

jmbarbone:
That makes sense.  When I converted all the `ReferenceClass` objects to `{R6}` I wasn't checking if they could be simplified.  Wasn't sure how each object was being used.  I think `S3` would make a bit more sense and be a bit more lightweight.

R/class-color.R
```diff
@@ -0,0 +1,84 @@
...
+
+    #' @field tint tint
+    tint = NULL,
+
+    #' @description
+    #' Creates a new `wbColor` object
+    #' @param name A name of a color known to R
+    #' @param auto A boolean.
+    #' @param indexed An indexed color values.
+    #' @param rgb A rgb color as ARGB hex value "FF000000".
```
jmbarbone, line 31: Maybe `hex` as the param name instead of `rgb`? I'd imagine we could have `rgb` to be a 3-length vector of integers and check for a `name`, `hex`, and `rgb` value as possible entries.

R/class-color.R
```diff
@@ -0,0 +1,84 @@
+#' R6 class for a Workbook Color
+#'
+#' A color
+#'
+#' @export
+wbColor <- R6::R6Class(
```
jmbarbone, line 6: As discussed, an `S3` object may be easy to do.  Can tag the `class`es as `openxlsx2` and `wbColor` for class checking.

R/class-color.R
jmbarbone: If all are `NULL` can we default to `name = "black"`?  It might be more simple to call `wb_color()` as the defaults for other arguments in functions.

R/wb_styles.R
```diff
@@ -873,12 +877,12 @@ set_cell_style <- function(wb, sheet, cell, value) {
 create_dxfs_style <- function(
     font_name      = "Calibri",
     font_size      = "11",
-    font_color     = c(rgb = "FF9C0006"),
+    font_color     = wb_color(rgb = "FF9C0006"),
```
jmbarbone, line 880: Do we want to use a default font color in `options()` like we have for `options(openxlsx2.borderColor)`?

R/wb_styles.R
```diff
@@ -873,12 +877,12 @@ set_cell_style <- function(wb, sheet, cell, value) {
...
     numFmt         = NULL,
     border         = NULL,
-    border_color   = getOption("openxlsx2.borderColour", "black"),
+    border_color   = wb_color(getOption("openxlsx2.borderColour", "black")),
```
jmbarbone, line 883: _color_ not _colour_?  We may want to check for the _u_ spelling anywhere else.

jmbarbone (changes requested):

R/wb_styles.R
JanMarvin: Yeah, I'll create a search & replace PR for that, not gonna write deprecate functions for all these changes.

R/wb_styles.R
JanMarvin: I'm a bit undecided. 
1. I really really dislike how the function calls look, if they have all the `options(openxlsx2.xxx)` calls. Changing things under the hood, if not `NULL` seems like a bad idea too. 
2. Maybe we could go with a `name = "default"` arg. That will call the `options(openxlsx2.borderColor)` under the hood? But than again, we'd have to make sure that everywhere a "default" option is possible. That easily turns into a plethora of changes.

R/class-color.R
JanMarvin: Regarding the first, in open xml hex values are `<color rgb="..."/>`, therefore it seems natural, but I agree, calling it hex on our end might be a better idea. Allowing rgb values as numeric vector of three it not something I'm really interested in.

default black might be a good idea, might as well provide a default color option.

JanMarvin:
Looong rebase. 

* Renamed it to `wb_colour()`. If CI is fine it'll get a NEWS bulletin and a merge. 
* The color/colour thing (#307) can wait. This will need a few more changes, to provide color and colour functions and arguments for every function.
* For now it's just a simple helper getting the needed xml color definitions. Therefore no breakage.

If we like this approach, we can go deeper in the future, prepare default options and asserts. But that's for future releases/rainy days.

JanMarvin:
There are a few inherits, that can be removed. Otherwise we enforce `wb_colour()` which was my idea some time ago.

JanMarvin:
well ... needs more fixes. 😞

PR #306, opened 2022-08-09
style picking

JanMarvin:
Provides `wb_get_cell_style()` and `wb_set_cell_style()`.

```R
wb <- wb_load(system.file("extdata", "oxlsx2_sheet.xlsx", package = "openxlsx2"))
styles <- wb_get_cell_style(wb, 1, "A10:G15")
wb <- wb_set_cell_style(wb, 1, "A30:G35", styles)
wb$open()
```

Needs
* [x] wrapper test
* [x] test
* [x] vignette entry
* [ ] copy roxygen


JanMarvin:
@jmbarbone fyi: I'm not entirely happy with the return of `wb_get_cell_style()`. It simply returns the style id as character. This makes sense in the way `openxlsx2` handles styles, but (a) it breaks the flow, chaining is not possible this way, and (b) we might want to make this a class we can check for and provide a custom print model, so that the user wont be entirely confused when seeing the results and debugging chains or pipes.

We might want to look at this for possible improvements in the future, but for this pull request and most likely for the upcoming release I'll go with the implemented approach. It works and is a slight improvement over the current status quo. That's currently good enough for me.

PR #267, opened 2022-07-10
[WIP] check for missing files

JanMarvin:
Start work on #218 

* [x] check `[Content_Types].xlsx`
* [ ] check various `_rels` directories. Not every entry is a directory, therefore need to check for e.g. hyperlinks as well. Might circle through `xl` sub-directories.
  * [ ] `/_rels`
  * [x] `/xl/_rels`
  * [x] `/xl/charts/_rels`
  * [x] `/xl/chartsheets/_rels`
  * [x] `/xl/drawings/_rels`
  * [x] `/xl/tables/_rels`
  * [x] `/xl/worksheets/_rels`
* [ ] remove entries from `[Content_Types].xlsx` and `_rels` if not available in output file
* [ ] add tests
* [ ] check for duplicates as well

PR #311, opened 2022-09-04
No drawings dir

JanMarvin:
This PR avoids creating drawings and vmlDrawings in every relationship header we create.

Previously, when we created workbook objects or worksheets, they referenced via `genBaseSheetRels()` to 
* `drawings` and
* `vmlDrawings`.

This caused problems in some third-party programs such as SPSS and Power BI (https://github.com/ycphs/openxlsx/issues/231). These programs seem to be more strict about the files available. They try to follow every reference and complain if a file is not found (after all, we did not supply these files). This makes them stricter than the usual spreadsheet software like Excel, LibreOffice or Numbers.

However, with this PR, our previous behavior is changed. We now only reference files that we actually deliver. If we create a drawing, or if the file we import contains a drawing, we will create drawings and generate the appropriate references. But if we haven't seen any of those files, we don't deliver them. This is the remaining fix 1 from https://github.com/ycphs/openxlsx/issues/231#issuecomment-907809082.
Ideally we will review this once we have created a file, but #267 is still early WIP and will not make it into the next release.

```R
> library(openxlsx2)
> 
> temp <- temp_xlsx()
> export_df <- wb_workbook()$add_worksheet("Sheet 1")$add_data(x = mtcars)$save(temp)
> # no reference to drawings
> unzip(temp, exdir = tempdir())
> read_xml(paste0(tempdir(), "/[Content_Types].xml"))
<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">
 <Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml" />
 <Default Extension="xml" ContentType="application/xml" />
 <Override PartName="/docProps/app.xml" ContentType="application/vnd.openxmlformats-officedocument.extended-properties+xml" />
 <Override PartName="/docProps/core.xml" ContentType="application/vnd.openxmlformats-package.core-properties+xml" />
 <Override PartName="/xl/styles.xml" ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.styles+xml" />
 <Override PartName="/xl/workbook.xml" ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet.main+xml" />
 <Override PartName="/xl/theme/theme1.xml" ContentType="application/vnd.openxmlformats-officedocument.theme+xml" />
 <Override PartName="/xl/worksheets/sheet1.xml" ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.worksheet+xml" />
</Types>
> # no _rels directory at all
> dir(paste0(tempdir(), "/xl/worksheets/"))
[1] "sheet1.xml"
```

In addition it is now possible to assign comments to sheets with vml objects:
```R
wb <- wb_load("https://github.com/JanMarvin/openxlsx-data/raw/main/macro2.xlsm")
c1 <- create_comment(text = "this is a comment", author = "")
write_comment(wb, 1, col = "B", row = 10, comment = c1)
```

### DIFF
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -5289,14 +5289,12 @@ wbWorkbook <- R6::R6Class(
           cd <- unapply(self$comments[[i]], "[[", "clientData")
           nComments <- length(cd)
 
-          vml_comment <- '<o:shapelayout v:ext="edit"><o:idmap v:ext="edit" data="1"/></o:shapelayout>'
+          vml_comment <- '<o:shapelayout v:ext="edit"><o:idmap v:ext="edit" data="1"/></o:shapelayout><v:shapetype id="_x0000_t202" coordsize="21600,21600" o:spt="202" path="m,l,21600r21600,l21600,xe"><v:stroke joinstyle="miter"/><v:path gradientshapeok="t" o:connecttype="rect"/></v:shapetype>'
 
           for (j in seq_len(nComments)) {
             id <- id + 1L
             vml_comment <- c(
-              vml_comment,
-              paste0('<v:shapetype id="_x0000_t202" coordsize="21600,21600" o:spt="202" path="m,l,21600r21600,l21600,xe"><v:stroke joinstyle="miter"/><v:path gradientshapeok="t" o:connecttype="rect"/></v:shapetype>',
-              genBaseShapeVML(cd[j], id))
+              vml_comment, genBaseShapeVML(cd[j], id)
             )
           }
         }
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -48,8 +48,8 @@ wbWorkbook <- R6::R6Class(
     #' @field drawings_rels drawings_rels
     drawings_rels = NULL,
 
-    #' @field drawings_vml drawings_vml
-    drawings_vml = NULL,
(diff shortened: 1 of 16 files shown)


R/class-workbook.R
```diff
@@ -2402,19 +2456,25 @@ wbWorkbook <- R6::R6Class(
...
         vmlDrawing_id <- xml_rels$target_ind[xml_rels$type == "vmlDrawing"]
 
         # NULL the sheets
         if (length(comment_id))    self$comments[[comment_id]]            <- NULL
-        if (length(drawing_id))    self$drawings[[drawing_id]]            <- NULL
-        if (length(drawing_id))    self$drawings_rels[[drawing_id]]       <- NULL
+        # if (length(drawing_id))
+                                   self$drawings[[sheet]]                 <- NULL
+        # if (length(drawing_id))
+                                   self$drawings_rels[[sheet]]            <- NULL
```
JanMarvin, line 2472: Needs to select drawing sheet from worksheets_rels. Sheet no != drawing no.

JanMarvin: drawing_id should be enough

JanMarvin:
Merged, but this needs close tracking

PR #326, opened 2022-09-12
align dims and rows & cols

JanMarvin:
if both are provided, they should be identical

### DIFF
--- a/R/wb_functions.R
+++ b/R/wb_functions.R
@@ -279,12 +279,6 @@ wb_to_df <- function(
 
   if (!is.null(cols)) cols <- col2int(cols)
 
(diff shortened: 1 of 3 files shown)


JanMarvin:
back to the drawing board. there is no guarantee that rows or cols are continuous sequences.

R/wb_functions.R
```diff
@@ -609,8 +632,8 @@ wb_to_df <- function(
 
     if (any(empty)) {
       sel <- which(names(empty) %in% names(empty[empty == TRUE]))
-      z[sel]  <- NULL
-      tt[sel] <- NULL
+      z[, sel]  <- NULL
+      tt[, sel] <- NULL
```
JanMarvin, line 636: revert this

Issue #20, opened 2022-01-01
`options()` tracking


jmbarbone:
Organize and track the potential default options. Most of these actually seem to be related to styles.

In `openxlsx` these were saved to the named list `op.openxlsx` and documented (although in the package document, `openxlsx`).

- `bandedCols`
- `bandedRows`
- `borderColour`
- `borders`
- `borderStyle`
- `compressionLevel`
- `creator` (`author`?)
- `dateFormat`, `dateformat` (just keep one)
- `datetimeFormat`, `datetimeformat`, `dateTimeFormat` (just keep one)
- `hdpi`
- `header`
- `headerStyle`
- `firstColumn`
- `firstFooter`
- `firstHeader`
- `footer`
- `evenFooter`
- `evenHeader`
- `gridLines`
- `keepNA`
- `lastColumn`
- `na.string`
- `maxWidth`
- `minWidth`
- `numFmt`
- `oddFooter`
- `oddHeader`
- `orientation`
- `paperSize`
- `showGridLines`
- `tabColour`
- `tableStyle`
- `vdpi`
- `withFilter`

<details>
<summary> From `openxlsx::op.openxlsx` </summary>

```r
list(
  openxlsx.bandedCols       = FALSE,
  openxlsx.bandedRows       = TRUE,
  openxlsx.borderColour     = "black",
  openxlsx.borders          = NULL,
  openxlsx.borderStyle      = "thin",
  openxlsx.compressionLevel = 9,
  openxlsx.creator          = "",
  openxlsx.dateFormat       = "mm/dd/yyyy",
  openxlsx.datetimeFormat   = "yyyy-mm-dd hh:mm:ss",
  openxlsx.hdpi             = 300,
  openxlsx.header           = NULL,
  openxlsx.headerStyle      = NULL,
  openxlsx.firstColumn      = NULL,
  openxlsx.firstFooter      = NULL,
  openxlsx.firstHeader      = NULL,
  openxlsx.footer           = NULL,
  openxlsx.evenFooter       = NULL,
  openxlsx.evenHeader       = NULL,
  openxlsx.gridLines        = TRUE,
  openxlsx.keepNA           = FALSE,
  openxlsx.lastColumn       = NULL,
  openxlsx.na.string        = NULL,
  openxlsx.maxWidth         = 250,
  openxlsx.minWidth         = 3,
  openxlsx.numFmt           = "GENERAL",
  openxlsx.oddFooter        = NULL,
[...]

PR #331, opened 2022-09-17
329 sheet name

### LINKED ISSUE #329: Unintended whitespace in waiver
https://github.com/JanMarvin/openxlsx2/blob/86c910f7da32f087a0c07f67886fd97be56e7238/R/class-workbook.R#L376

This creates "Sheet  1" instead of "Sheet 1". Just for reference Excel uses "Sheet1". In addition this is localized, in German Excel sheets are called "Tabelle" instead of Sheet. If we ever run out of ideas what to do, we could provide something like this too.

jmbarbone:
resolves #329 

- remove whitespace
- add sheet name to option list
- redoc

### DIFF
--- a/R/class-workbook-wrappers.R
+++ b/R/class-workbook-wrappers.R
@@ -2026,7 +2026,7 @@ wb_add_style <- function(wb, style = NULL, style_name = NULL) {
 #'
 #' # get style from b1 to assign it to a1
 #' numfmt <- wb$get_cell_style(dims = "B1")
-#' 
+#'
 #' # assign style to a1
 #' wb$set_cell_style(dims = "A1", style = numfmt)
 #' @return wb_get_cell_style returns the style id as character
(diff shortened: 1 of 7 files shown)


jmbarbone:
@JanMarvin , I had created an issue long ago (#20) to make these options a bit more visible and well documented (and then I didn't get very far).  Something like this can be handled pretty easily and I don't think we'd run into many issues allowing users to specify the default sheet name.

JanMarvin (approved):

JanMarvin:
Thanks, nice idea! Just one more thing, if I remember correctly there are a few places where I have seen `Sheet  #` (with two whitespaces), maybe in vignettes or examples. Could you check if something pops up with `s/Sheet  /Sheet /g`?

JanMarvin:
@jmbarbone , I've pushed a minor modification switching paste for paste0. This way it's possible to use "Sheet" to avoid having whitespace in the first place.
If you're fine with this, please merge.

PR #341, opened 2022-09-26
change default col size to 10

JanMarvin:
This adjusts a few functions to use default column width of 10 and default font size 12. This is what MS365 uses on Mac. It is changed, because otherwise column widths were slightly off. 9. something instead of 10. Adjusted to this new values, all calculations are correct for Calibri. Yet, I would not bet that our calculations are valid for all versions. Previously we were using 8.43 and font size 11 as default. Which might be correct for the MS365 webbrowser edition or certain Windows low dpi versions.

Maybe you can check the following on Windows @jmbarbone ? On Mac the column width is exactly 8.5 ... 17.5.
```R
  wb <- wb_workbook()$
    add_worksheet()$
    set_col_widths(cols = 1:10, width = (8:17) + .5)$
    add_data(x = rbind(8:17), colNames = FALSE)$
    open()
```

### DIFF
--- a/R/baseXML.R
+++ b/R/baseXML.R
@@ -236,7 +236,7 @@ genBaseStyleSheet <- function(dxfs = NULL, tableStyles = NULL, extLst = NULL) {
   list(
     numFmts = NULL,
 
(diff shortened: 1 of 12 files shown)


JanMarvin:
This is still a pending pull request (I've reverted it, needs a reland if tests come up fine)

PR #350, opened 2022-09-30
[reland] cleanup column width and font size setting

JanMarvin:
I have updated to font size settings to default to column width of 8.43 and a font size of 11 and to use the windows size setting. On Mac the column width will be slightly off, but otherwise we'd have to skip a couple of tests.

~~This adjusts a few functions to use default column width of 10 and default font size 12. This is what MS365 uses on Mac. It is changed, because otherwise column widths were slightly off. 9. something instead of 10. Adjusted to this new values, all calculations are correct for Calibri. Yet, I would not bet that our calculations are valid for all versions. Previously we were using 8.43 and font size 11 as default. Which might be correct for the MS365 webbrowser edition or certain Windows low dpi versions.~~

Maybe you can check the following on Windows @jmbarbone ? On Mac the column width is exactly 8.5 ... 17.5.

```R
  wb <- wb_workbook()$
    add_worksheet()$
    set_col_widths(cols = 1:10, width = (8:17) + .5)$
    add_data(x = rbind(8:17), colNames = FALSE)$
    open()
```

Orignal PR was #341 which was accidentally merged and reverted

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,7 +14,7 @@
 
 ## Fixes
 
-* Improvements to setting column widths. Previously values set by `set_col_widths()` were a little off. This has been fixed. [350](https://github.com/JanMarvin/openxlsx2/pull/350)
+* Improvements to setting column widths. Previously values set by `set_col_widths()` were a little off. This has now been improved. There are still corner cases where the column width set with `openxlsx2` does not match those shown in spreadsheet software. Notable differences can be seen with floating point values (e.g., `10L` works while `10.1` is slightly off) and with column width on Mac. [350](https://github.com/JanMarvin/openxlsx2/pull/350)
 
 * Improve `rowNames` when writing data to worksheet. Previously the name for the rownames column defaulted to `1`. This has been changed. Now with data it defaults to an empty cell and with a data table it defaults to `_rowNames_`. [375](https://github.com/JanMarvin/openxlsx2/pull/375)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,8 @@
 
 ## Fixes
 
+* Improvements to setting column widths. Previously values set by `set_col_widths()` were a little off. This has been fixed. [350](https://github.com/JanMarvin/openxlsx2/pull/350)
+
 * Improve `rowNames` when writing data to worksheet. Previously the name for the rownames column defaulted to `1`. This has been changed. Now with data it defaults to an empty cell and with a data table it defaults to `_rowNames_`. [375](https://github.com/JanMarvin/openxlsx2/pull/375)
 
 * Fix the workbook xml relationship file to not include a reference to shared strings per default. This solves [360](https://github.com/JanMarvin/openxlsx2/issues/360) for plain data files written from `openxlsx2`. [363](https://github.com/JanMarvin/openxlsx2/pull/363)
--- a/R/converters.R
+++ b/R/converters.R
@@ -61,3 +61,40 @@ get_cell_refs <- function(cellCoords) {
   l <- int2col(unlist(cellCoords[, 2]))
   paste0(l, cellCoords[, 1])
 }
+
+
+
+#' calculate the required column width
+#'
+#' @param base_font the base font name and fontsize
+#' @param col_width column width
+#' @keywords internal
+#' @examples
+#' base_font <- wb_get_base_font(wb)
+#' calc_col_width(base_font, col_width = 10)
+#' @noRd
+calc_col_width <- function(base_font, col_width) {
+
+  # TODO save this instead as internal package data for quicker loading
+  fw <- system.file("extdata", "fontwidth/FontWidth.csv", package = "openxlsx2")
+  font_width_tab <- read.csv(fw)
+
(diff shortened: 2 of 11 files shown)


jmbarbone (changes requested):
@JanMarvin, I'm seeing an extra `0.25` on Windows.  Column widths are appearing as `8.75`, `9.75`, etc.

JanMarvin:
Thanks @jmbarbone , I have modified their formula here:

```R
widths <- trunc((as.numeric(col_width) * mdw + 5 + 2) / mdw * 256) / 256
```

The `+2` is my fix for Mac. Could you see if it works without the `+2`? I'm not sure if the width relies on the OS or DPI of the screen somehow. I will check with the web browser version of MS365.

jmbarbone:
@JanMarvin , removing the `+2` seems to do the trick.  All the columns are at the appropriate `x.5` widths.

JanMarvin:
So their formula actually works on Windows. Maybe we should stick with that.
My initial hope was that the value in the XML file would match the value Excel shows. Though that is not the case and the value we are currently setting was off when applied on Mac. This PR is an improvement, though the default width seems to be varying across Excel and OS versions. MS365 on Mac has a default width of 10. My work Excel 2013 has 10.71 and MS365 in the browser was 8.43. Do you have any preference? (The calculated value is depending on the default width as well.)
We could go with the 10 and this PR and formula for Windows. This way if someone sets a column width the result would match on one OS.

JanMarvin:
@jmbarbone If you find the time, could you please give this another feature test and run the check again? Just want to make sure that we set the column width correctly for Windows Excel. Unfortunately the width values are different on Mac and I haven't found a way to please both Excels. Might be a general issue: [link to ms forum](https://answers.microsoft.com/en-us/msoffice/forum/all/different-column-widths-in-excel-for-windows-vs/ba5c4e90-7b27-4197-a027-9b79f9ae28c0)

This changes the column width setting to calculate the values, whereas previously we simply set a value which was not identical to whatever Excel was using. This aligns a few functions to return width 8.43 and Calibri 11 font. I'd like to include this with 0.3.1 due to the more expected column widths. Prior to this PR you would set the width to 10 and would end up with something different, now you end up with column width 10 on Windows and 9.8 or so on Mac.

jmbarbone:
@JanMarvin , with the current build the column widths seem fine when set to integers or integer-like values.  Decimals seem to have some random offset:

``` r
devtool::load_all()
wb <- wb_workbook()
wb$add_worksheet()
wb$add_data(x = rbind(8:17), colNames = FALSE)
wb$set_col_widths(cols = 1:10, width = (8:17) + seq(0.1, 1, .1))
wb$open()
```


| width| offset| actual| difference|
|-----:|------:|------:|----------:|
|     8|    0.1|   8.14|       0.04|
|     9|    0.2|   9.14|      -0.06|
|    10|    0.3|  10.29|      -0.01|
|    11|    0.4|  11.43|       0.03|
|    12|    0.5|  12.43|      -0.07|
|    13|    0.6|  13.57|      -0.03|
|    14|    0.7|  14.71|       0.01|
|    15|    0.8|  15.86|       0.06|
|    16|    0.9|  16.86|      -0.04|
|    17|    1.0|  18.00|       0.00|

JanMarvin:
Thanks. Might be related to the use of round. With .2 it is -0.06 and with .8 it's 0,06. Do you feel like this is reason enough to hold the pull request back? I'd say we can merge it the way it is and improve from there on (if we ever have to). After all it is already much better than the previous implementation.

R/converters.R
```diff
@@ -61,3 +61,40 @@ get_cell_refs <- function(cellCoords) {
...
+
+  # formula from openxml.spreadsheet.column documentation. The formula returns exactly the expected
+  # value, but the output in excel is still off. Therefore round to create even numbers. In my tests
+  # the results were close to the initial col_width sizes. Character width is still bad, numbers are
+  # way larger, therefore characters cells are to wide. Not sure if we need improve this.
+
+  # Note: cannot reproduce the exact values with MS365 on Mac. Nevertheless, these values are closer
+  # to the expected widths
+  widths <- trunc((as.numeric(col_width) * mdw + 5) / mdw * 256) / 256
+  widths <- round(widths, 3)
```
JanMarvin, line 98: This is the round I was thinking about

jmbarbone:
@JanMarvin , if this is an improvement from the current version, then it's probably good to include it

Might just need a disclaimer:

> column width setting has been improved but may not be exact on Windows

NEWS.md
```diff
@@ -14,6 +14,8 @@
 
 ## Fixes
 
+* Improvements to setting column widths. Previously values set by `set_col_widths()` were a little off. This has been fixed. [350](https://github.com/JanMarvin/openxlsx2/pull/350)
```
jmbarbone, line 17: Should note that it still may not be exact on Windows?

jmbarbone (approved):

NEWS.md
JanMarvin: Yes, will do

JanMarvin: Updated NEWS.

PR #358, opened 2022-10-03
change update_cell() logic. fixes #356

### LINKED ISSUE #356: Writing a large dataframe on a preformated worksheet is slow
While testing openxlsx2 I ran into a significant performance difference. In openxlsx writing ~10K rows to a sheet in a preformatted (but blank) workbook is quite quick. It's not instant but pretty close. In openxlsx2 this step takes 2.5-3 minutes. 

Unfortunately I'm not able to share the actual data or workbook but based on my tests I think the issue is the pre-existing formatting and conditional formatting. Creating a new workbook object and dumping 10K or more rows of dummy data is nearly instant, doing the same with the preformatted workbook loaded off disk takes minutes.

JanMarvin:
Currently an idea. 

A test is this:

```R
library(openxlsx2)

mm <- matrix(1,10000, 100)

wb <- wb_workbook()$
  add_worksheet()$
  add_data(x = mm)

wb$add_data(x = mm)
```

* [x] check if this is a valid replacement
* [x] add tests
* [x] remove dead code
* [x] remove `update_cell()` examples
* [x] creation and applying styles should only happen if there's no data on the sheet or if replace style is explicitly called


JanMarvin:
If nothing comes up, I'll merge this tomorrow. We can test and tweak this afterwards if we need to. The benefits of having it even though it might not yet be perfect outweigh potential issues.

PR #362, opened 2022-10-06
[wb_load] xml cleanup

JanMarvin:
* cleanup of baseXML
* fix opening workbooks with slicer
* fix cloning pivot charts

Unfortunately I didn't check loading sheets with and without slicers or I misinterpreted the errors I saw when cloning slicers. 

```R
# clone sheet with pivotTable (and drop slicer)
wb <- wb_load(file = system.file("extdata", "loadExample.xlsx", package = "openxlsx2"))$
  clone_worksheet(old = "irisSample", new = "clone 1")$
  open()
```

Issue #360, opened 2022-10-05
openxlsx2 ships invalid workbook.xml rel to sharedStrings.xml


rowrowrowyourboat:
If I generate a file using openxlsx2 and then try to connect to it from Excel, Excel says that the file is not a valid Excel document. I'm connecting to it in Excel using these menu steps: Data > New Query > From File > From Excel workbook.

Excel error message:
# We encountered an error while trying to connect.
# Details: "The input couldn't be recognized as a valid Excel document."
# Buttons: Retry/Edit/Cancel

I can open that file in Excel just fine. If I open it in Excel and then save it from Excel, I can then connect to the file.

I'm using Microsoft Office Professional Plus 2016, Windows 11, R 4.2.1 and openxlsx2 0.3 (more details in reprex below)

I'm attaching the file generated by openxlsx2 and that same file after it was opened and saved by Excel.
[mtcars_savedbyExcel.xlsx](https://github.com/JanMarvin/openxlsx2/files/9717871/mtcars_savedbyExcel.xlsx)
[mtcars.xlsx](https://github.com/JanMarvin/openxlsx2/files/9717872/mtcars.xlsx)

Here the code to reproduce it:

``` r
remotes::install_github("JanMarvin/openxlsx2")
#> Skipping install of 'openxlsx2' from a github remote, the SHA1 (739218a3) has not changed since last install.
#>   Use `force = TRUE` to force installation

library(openxlsx2)

print(sessionInfo())
#> R version 4.2.1 (2022-06-23 ucrt)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 10 x64 (build 22000)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United States.utf8 
#> [2] LC_CTYPE=English_United States.utf8   
#> [3] LC_MONETARY=English_United States.utf8
#> [4] LC_NUMERIC=C                          
#> [5] LC_TIME=English_United States.utf8    
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] openxlsx2_0.3.0.90000
#> 
#> loaded via a namespace (and not attached):
#>  [1] zip_2.2.1         Rcpp_1.0.9        pillar_1.7.0      compiler_4.2.1   
[...]

PR #363, opened 2022-10-06
Gh issue 360

JanMarvin:
Avoid adding `sharedStrings.xml` to `workbook.xml.rels`. This is just a patch to solve issue #360. The better fix would be to not mess with `workbook.xml.rels` at all when loading. Though that requires a deeper look at the code. This is based on #362.

### DIFF
--- a/R/wb_load.R
+++ b/R/wb_load.R
@@ -133,15 +133,16 @@ wb_load <- function(file, xlsxFile = NULL, sheet, data_only = FALSE) {
 
     # TODO we remove all of the relationships and add them back later on.
     # I am not sure why we do this. We should simply keep it the way it is,
-    # and if we need to.
+    # and only modify it if we need to.
 
-    # Currently create a slim skeleton and add relationships if we see one. An
-    # alternative are the snippets below. Though this leaves the possibilty
-    # that we still keep folders we do not handle. Like ctrlProps #206.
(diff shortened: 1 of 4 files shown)


JanMarvin:
For now this is a hackish fix to solve #360. Unfortunately this might break any time a more complex sheet is required.

* ideally we should load the original relationship and update when needed, or
* we could create a new relationship file when saving

I currently prefer the second approach, but we'd have to research what precisely goes into this. Currently I do not have the time for this, hence this PR.

R/wb_load.R
```diff
@@ -121,10 +127,46 @@ wb_load <- function(file, xlsxFile = NULL, sheet, data_only = FALSE) {
...
+    # I am not sure why we do this. We should simply keep it the way it is,
+    # and if we need to.
+
+    # Currently create a slim skeleton and add relationships if we see one. An
+    # alternative are the snippets below. Though this leaves the possibilty
+    # that we still keep folders we do not handle. Like ctrlProps #206.
+
+    # Ideally we keep the entire relationship as loaded and check only for
+    # unknown content. Lastly workbook.xml.rels should be checked for
+    # consistency pre or post writing the output.
```
JanMarvin, line 144: This comment almost makes sense.  Keep it up!

PR #367, opened 2022-10-09
remove deprecated. fixes #239

### LINKED ISSUE #239: Rgui names deprecated warning
When importing a workbook in Rgui and selecting `wb$sharedStrings` I see a warning that I should use `get_sheet_names`. Not sure if spurious or if we use `names` somewhere in our code.

JanMarvin:
This caused spurious warnings in RGui and Visual Studio Code.

@jmbarbone let me know what you think about this (iirc you deprecated `names.wbWorkbook`). In this PR I simply removed the functions altogether, but we could keep them without the deprecation warning (everything else, printing a message will probably pollute the user console as well). The warnings were most likely thrown by IDEs due to `names()` being such an incremental part of R.
In addition, should we provide a `names.wbWorkbook`? With this PR `names(wb)` will print all the internal function names.

On a side note, we probably should think about a time span for deprecated functions. Do we keep them around until someone finally removes them or for the next dot release or for a time span of x days/weeks/months?


jmbarbone (approved):

jmbarbone:
Looks fine.  I guess my use of _deprecated_ here is a little weird since I did that _before_ it was CRAN released.  I think it's fine to trim out any of that stuff right now.

I don't see much of a utility in `names.wbWorkbook()` right now.  I feel like there are a lot of potential "names" and it might be better to have specific methods like what we have for sheets (`$get_sheet_names()`).  It's still an `{R6}` object so the default behavior seems fine.

For future deprecating, I agree, that we should probably wait at least until the next minor release to do so, and note that in the warning message.

PR #395, opened 2022-10-29
release 0.3.1

JanMarvin:
* 2022-10-30: First submission 
Two mails regarding auto-check sprintf.


JanMarvin:
The CRAN release is currently being held due to this being found in our log: `Rcpp/internal/r_coerce.h:255:7: warning: 'sprintf' is deprecated`. I have replied to CRAN that this is not caused by our package and already [fixed upstream](https://github.com/RcppCore/Rcpp/pull/1236).
I assume that the reason it ends up in our log is due to our usage of `Rcpp::String()`. In the worst case we might have to wait a couple of days for the next Rcpp release, but haven't heard back yet from CRAN.

PR #399, opened 2022-11-03
Set sheetview

JanMarvin:
Provide worksheet function `set_sheetview()`. This function can be used to set and modify attributes. This function can be wrapped in a wbWorkbook function (and moved to private later on).
In addition it provides a new option to set `rightToLeft` as default for new worksheets.

* [x] wbChartSheet has a sheetview as well
* [x] Excel test is pending
* [x] update NEWS


JanMarvin:
There is an unrelated bug in charts vml handling:

```R
library(openxlsx2)

fl <- system.file("extdata", "mtcars_chart.xlsx", package = "openxlsx2")
wb <- wb_load(fl)
wb$save("/tmp/test.xlsx")
# wb$open() # broken vmlDrawing1 gets vmlDrawing4

unlink("/tmp/xlsx", recursive = TRUE)
unlink("/tmp/test", recursive = TRUE)

unzip(fl, exdir = "/tmp/xlsx")
unzip("/tmp/test.xlsx", exdir = "/tmp/test")
```

R/class-workbook.R
```diff
@@ -5895,24 +5923,13 @@ wbWorkbook <- R6::R6Class(
...
-            )
-        }
-      }
+      # # TODO what does this code segment do?
+      # self$worksheets[[visible_sheet_index]]$set_sheetview(tabSelected = TRUE)
+      # if (nSheets > 1) {
+      #   for (i in setdiff(seq_len(nSheets), visible_sheet_index)) {
+      #     self$worksheets[[i]]$set_sheetview(tabSelected = FALSE)
+      #   }
+      # }
```
JanMarvin, line 5932: Maybe this is actually a required failsafe. Hidden sheets should not be tabSelected.

R/class-chart-sheet.R
```diff
@@ -65,24 +51,99 @@ wbChartSheet <- R6::R6Class(
...
-#' New chart sheet
-#'
-#' Create a new chart sheet
-#'
-#' @returns a `wbChartSheet` object
-#' @export
-wb_chart_sheet <- function() {
-  wbChartSheet$new()
-}
+)
```
JanMarvin, line 149: needs a new line and this wb function was never used, still should be mentioned in NEWS

PR #403, opened 2022-11-05
[wb_load] update vml file name. fixes #402

### LINKED ISSUE #402: Update vmlDrawing in worksheets_rels
There is an unrelated bug in charts vml handling (in this case it is related to embedded files, therefore most likely not that visible to users):

```R
library(openxlsx2)

fl <- system.file("extdata", "mtcars_chart.xlsx", package = "openxlsx2")
wb <- wb_load(fl)
wb$save("/tmp/test.xlsx")

wb$worksheets_rels[[4]][2] <- # vmlDrawing is now vmlDrawing4 not 1
  "<Relationship Id=\"rId2\" Type=\"http://schemas.openxmlformats.org/officeDocument/2006/relationships/vmlDrawing\" Target=\"../drawings/vmlDrawing4.vml\"/>" 
  
wb$open()

unlink("/tmp/xlsx", recursive = TRUE)
unlink("/tmp/test", recursive = TRUE)

unzip(fl, exdir = "/tmp/xlsx")
unzip("/tmp/test.xlsx", exdir = "/tmp/test")
```

JanMarvin:
We update vmlDrawing name, but ignored it in worksheets_rels.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # openxlsx2 (in development)
 
+## Fixes
+
+* Fixed a case where embedded files were assigned incorrectly in worksheet relationships. This caused corrupted output. [403](https://github.com/JanMarvin/openxlsx2/pull/403)
 
 ***************************************************************************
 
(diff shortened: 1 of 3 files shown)


JanMarvin:
I do not really like this. It is a hackish approach to fix the issue. One day we should really fix this in a saner way. Like auto generate the relationship files when saving or something alike.

PR #405, opened 2022-11-06
[wb_load] read x namespace

JanMarvin:
allow reading files with x namespace

## what was the motivation for this PR?

This fixes import for files created by third party software as seen in https://github.com/ycphs/openxlsx/issues/344.

Maybe one of you can test this @Matpli, @elinw? You can build this branch using:

```R
remotes::install_github("JanMarvin/openxlsx2", ref = "gh_issue_213")
```

## what does this PR do?

I have taken the initial #213 and created a few more safeguards to make sure that we do not have to check every xml file we import. 

1. Instead now we only check if we find `x:workbook` in `/xl/workbook.xml` 
1. If TRUE set `options("openxlsx2.has_x_namespace" = TRUE)`
1. If this option is true we read the file into R
1. If this string is found `xmlns:x="http://schemas.openxmlformats.org/spreadsheetml/2006/main"`, we will remove the `x` namespace from the input
1. We let the user know that this is somewhat experimental and largely untested

It works with the example file provided by @Matpli and if I save this file I can open it in MS365. But this file is quite large so if anyone can provide a smaller example file, that would be great.

### DIFF
--- a/R/pugixml.R
+++ b/R/pugixml.R
@@ -65,7 +65,7 @@ read_xml <- function(xml, pointer = TRUE, escapes = FALSE, declaration = FALSE, 
       stringi::stri_read_lines(xml, encoding = "UTF-8"),
       collapse = "")
 
-    if (grepl('xmlns:x="http://schemas.openxmlformats.org/spreadsheetml/2006/main"', xml_file)) {
+    if (grepl('xmlns:x="http://schemas.openxmlformats.org/spreadsheetml/2006/main"', xml_file, fixed = TRUE)) {
       xml_file <- stringi::stri_replace_all_fixed(xml_file, "<x:", "<")
       xml_file <- stringi::stri_replace_all_fixed(xml_file, "</x:", "</")
 
--- a/R/pugixml.R
+++ b/R/pugixml.R
@@ -56,18 +56,18 @@ read_xml <- function(xml, pointer = TRUE, escapes = FALSE, declaration = FALSE, 
   if (identical(xml, ""))
     xml <- "<NA_character_ />"
 
-  # Clean xml files from x namespace. Otherwise all nodes might look like
+  # Clean xml files from xml namespace. Otherwise all nodes might look like
   # <x:node/> and not <node/>. https://github.com/JanMarvin/openxlsx2/pull/213
-  has_x_ns <- getOption("openxlsx2.namespace_x", default = FALSE)
-  if (has_x_ns && isfile && !isvml) {
+  xml_ns <- getOption("openxlsx2.namespace_xml")
+  if (!is.null(xml_ns) && isfile && !isvml) {
 
     xml_file <- stringi::stri_join(
       stringi::stri_read_lines(xml, encoding = "UTF-8"),
       collapse = "")
 
-    if (grepl('xmlns:x="http://schemas.openxmlformats.org/spreadsheetml/2006/main"', xml_file, fixed = TRUE)) {
-      xml_file <- stringi::stri_replace_all_fixed(xml_file, "<x:", "<")
-      xml_file <- stringi::stri_replace_all_fixed(xml_file, "</x:", "</")
+    if (grepl(sprintf('xmlns:%s="http://schemas.openxmlformats.org/spreadsheetml/2006/main"', xml_ns), xml_file, fixed = TRUE)) {
+      xml_file <- stringi::stri_replace_all_fixed(xml_file, sprintf("<%s:", xml_ns), "<")
+      xml_file <- stringi::stri_replace_all_fixed(xml_file, sprintf("</%s:", xml_ns), "</")
 
       # replace xml with already and cleaned output
       xml <- xml_file
--- a/R/pugixml.R
+++ b/R/pugixml.R
@@ -44,6 +44,8 @@ read_xml <- function(xml, pointer = TRUE, escapes = FALSE, declaration = FALSE, 
 
   z <- NULL
 
+  isvml <- grepl("^.vml$", xml)
+
   isfile <- FALSE
   if (length(xml) == 1 && file.exists(xml))
     isfile <- TRUE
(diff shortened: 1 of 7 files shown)


R/pugixml.R
```diff
@@ -54,6 +56,25 @@ read_xml <- function(xml, pointer = TRUE, escapes = FALSE, declaration = FALSE,
   if (identical(xml, ""))
     xml <- "<NA_character_ />"
 
+  # Clean xml files from x namespace. Otherwise all nodes might look like
+  # <x:node/> and not <node/>. https://github.com/JanMarvin/openxlsx2/pull/213
+  has_x_ns <- isTRUE(options()[["openxlsx2.has_x_namespace"]])
```
jmbarbone, line 61: `getOption()` instead

would `options(openxlsx2.namespace_x = TRUE)` make sense?

R/pugixml.R
```diff
@@ -54,6 +56,25 @@ read_xml <- function(xml, pointer = TRUE, escapes = FALSE, declaration = FALSE,
...
+  if (has_x_ns && isfile && !isvml) {
+
+    xml_file <- stringi::stri_join(
+      stringi::stri_read_lines(xml, encoding = "UTF-8"),
+      collapse = "")
+
+    if (grepl('xmlns:x="http://schemas.openxmlformats.org/spreadsheetml/2006/main"', xml_file)) {
```
jmbarbone, line 68: `grepl(fixed = TRUE)`?

R/wb_load.R
```diff
@@ -196,6 +196,17 @@ wb_load <- function(file, xlsxFile = NULL, sheet, data_only = FALSE) {
...
     workbook_xml <- read_xml(workbookXML, escapes = TRUE)
 
+    if (length(xml_node_name(workbook_xml, stringi::stri_join("x:workbook")))) {
+      op <- options("openxlsx2.has_x_namespace" = TRUE)
+      on.exit(options(op), add = TRUE)
+      out <- paste0(
+        "File has x namespace. We correct this for you, but be careful.\n",
+        "This is somewhat experimental.."
+      )
+      message(out)
```
jmbarbone, line 206: Should this be a `message` or a `warning`?  

Maybe a bit more information on how the x namespace is correct?  I'm not quite sure about what this is, and it could be a little confusing to see this message

jmbarbone (changes requested):
Adding in some information what the x namespace is could be helpful when people counter this.  I don't know what it is or any of the implications of changing these values are, so I'd certainly be confused.

R/pugixml.R
JanMarvin: Thanks wasn't aware of `getOption()`, will add this in #399

R/pugixml.R
JanMarvin: added

R/wb_load.R
JanMarvin: Tried to make this a bit more user friendly. Wasn't sure if warning or message, but warning might be a good idea. This way it will be easier to spot and we can always decrease this later

JanMarvin:
> Adding in some information what the x namespace is could be helpful when people counter this. I don't know what it is or any of the implications of changing these values are, so I'd certainly be confused.

The example file used in the test wraps `main` in this x namespace: `xmlns:x="http://schemas.openxmlformats.org/spreadsheetml/2006/main"`. Therefore all xml nodes are wrapped in this namespace. We - and Excel and every other spreadsheet software I'm aware of - use ` xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"` therefore we do not need a namespace prefix.

Our xml nodes look like this `<workbook>` and in the example file they look like this `<x:workbook>`. But `xml_node("<x:workbook/>", "workbook")` will not return the expected result. Once we remove the x namespace from the nodes, we can read the file. Therefore this PR should be safe when importing with x namespace. However, since we are changing the xml files before importing and we currently only have this one file to test, I wanted to be careful.

In theory this namespace does not have to be named x and could be something else, though I didn't really want to search for other namespaces. In openxml there are various namespaces like `x14` therefore assuming that it will be `x` should be safe. It is also mentioned in `ECMA-376 Part 1:  Annex D. Namespace Prefix Mapping in Examples`.

JanMarvin:
The more general fix would be to read the header and replace the namespace whenever they are found. 

But I wanted something that works quick and does not interfere with the overwhelming majority of files we import. If we get reports that users experience plenty of these warnings, we can have a look at the c++ loading functions of pugixml and see if we can implement a remove namespace wrapper there.

jmbarbone:
> In theory this namespace does not have to be named x and could be something else, though I didn't really want to search for other namespaces. In openxml there are various namespaces like `x14` therefore assuming that it will be `x` should be safe. It is also mentioned in `ECMA-376 Part 1: Annex D. Namespace Prefix Mapping in Examples`.

Maybe we can then use the options like:

``` r
options(openxlsx2.xml_namespace = "x") # maybe `TRUE` defaults to `"x"`?
```

or 

``` r
options(openxlsx2.xml_namespace = c("foo", "bar")))
```

then check to see if any have been specified?

R/wb_load.R
```diff
@@ -196,6 +196,17 @@ wb_load <- function(file, xlsxFile = NULL, sheet, data_only = FALSE) {
     # escape
     workbook_xml <- read_xml(workbookXML, escapes = TRUE)
 
+    if (length(xml_node_name(workbook_xml, stringi::stri_join("x:workbook")))) {
+      op <- options("openxlsx2.namespace_x" = TRUE)
+      on.exit(options(op), add = TRUE)
+      out <- paste0(
+        "This file was slightly modified to import it. It should be fine, but be careful.\n",
```
jmbarbone, line 203: Maybe something a bit more more specific?

``` 
The `{x}` namespace(s) has been removed from the xml files, for example:
  <x:field> changed to:
  <field>

See 'Details' in ?openxlsx2::wb_load() for more information.
```

jmbarbone (commented):
From some quick searches this morning it sounds like the default `x` namespace can be particular annoying and prevent other software from correctly parsing the xml.  It seems the go-to move is to remove the namespaces, so us doing it sounds right, especially if it's something the user has to opt into.

JanMarvin:
Haha, mate don't make this overly complicated for us. But your suggestions sound reasonable, I'll adjust the PR later.

JanMarvin:
Should be good to go now.

jmbarbone (approved):

PR #428, opened 2022-11-21
Enforce `fullCalcOnLoad`. closes #427

### LINKED ISSUE #427: Workbook formulas are not recalculated on opening in Excel
After amending an existing workbook in `openxlsx2`, formulas are not recalculating on opening the amended workbook in Excel. As a consequence, formula results remain wrong (and are quite difficult to get Excel to recalculate, too). I would argue that all workbooks saved by `openxlsx2::wb_save()` should be set to recalculate on opening in Excel.

To reproduce:
- Create a workbook in Excel, with a formula that relies on another cell in the workbook.
- Amend relied-upon value using `openxlsx2::wb_load()` and `openxlsx2::write_data()`, save via `openxlsx2::wb_save()`
- Open amended workbook in Excel: The formula shows the original result (pressing the Formulas > Calculation > Calculate Now button does not lead to a recalculation).

Screenshot 2022-11-21 104747

``` R
library(openxlsx2)

wb <- wb_load("test_calculation.xlsx")
write_data(
  wb = wb,
  sheet = "Sheet1",
  x = "Beta",
  startCol = 1,
  startRow = 1,
  colNames = FALSE
)
wb_save(wb, "test_calculation_result.xlsx")
```
[test_calculation.xlsx](https://github.com/JanMarvin/openxlsx2/files/10055767/test_calculation.xlsx)
[test_calculation_result.xlsx](https://github.com/JanMarvin/openxlsx2/files/10055768/test_calculation_result.xlsx)

JanMarvin:
If a formula in a workbook originally created by Excel refers to a cell modified by `openxlsx2`, the formula was not recalculated when loaded in Excel. The formula was update only if the cell was visited and Enter was hit. This was hard to spot could easily result in errors. This was solved enforcing [`fullCalcOnLoad`](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.calculationproperties.fullcalculationonload?view=openxml-2.8.1).

This also brings `revisionPtr` simply to reduce visual differences when comparing two files.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,10 @@
 
 * Fixed a case where embedded files were assigned incorrectly in worksheet relationships. This caused corrupted output. [403](https://github.com/JanMarvin/openxlsx2/pull/403)
 
+## Breaking changes
+
+* Previously if a loaded workbook contained formulas pointing to cells modified by `openxlsx2`, these formulas were not updated, once the workbook was opened in spreadsheet software. This is now enforced, unless the option `openxlsx2.disableFullCalcOnLoad` is set. In this case we would respect the original calculation properties of the  workbook.
+
 
 ***************************************************************************
 
--- a/R/baseXML.R
+++ b/R/baseXML.R
@@ -108,6 +108,7 @@ genBaseWorkbook <- function() {
     fileSharing = NULL,
     workbookPr = '<workbookPr date1904="false"/>',
     alternateContent = NULL,
+    revisionPtr = NULL,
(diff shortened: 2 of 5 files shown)


JanMarvin:
This should be correct, though it is not universally available. 
* Excel simply does nothing without it, works fine with it
* LibreOffice 7.4 seems to handle it partially. It works on a single sheet, but not with secondary sheets
* Numbers simply ignores the calculation properties, therefore no issue is visible

PR #431, opened 2022-11-22
378 wb save

### LINKED ISSUE #378: Avoid character returns in `wb` functions
Certain `wb` functions return character strings. Recently stumbled over this here:
``` r
library(openxlsx2)
wb <- wb_workbook() %>% 
  wb_add_worksheet() %>% 
  wb_save(path = temp_xlsx()) %>%
  wb_open()
#> Error: wb must be class wbWorkbook or R6
```
I expected `wb` to remain a workbook, but it had turned into a string with `wb_save()`.

 Another case where a `wb` function returns something other than a `wbWorkbook` is `wb_get_cell_styles()`:
``` r
library(openxlsx2)
wb <- wb_workbook() %>% 
  wb_add_worksheet() %>% 
  wb_get_cell_style(dims = "A1") %>% 
  wb_open()
#> Error: wb must be class wbWorkbook or R6
```

I'm wondering, if we should avoid this. I'm thinking about something like a path and message field, where we store the path we read from and save to.

jmbarbone:
resolves #378

I've added a replacement for the old behavior of wb_save(), which helps with some of the tests that used wb_save() and expected a path in the output.
I think that could make sense but I'm open to suggestions otherwise.

- correct behavior of wb_save() and include wb_write()
- redoc
- add test for wb_write()
- update wb_save() > wb_write() usage in tests
- update NEWS
- fix more usages of wb_save() > wb_write()


jmbarbone:
> Looking good, though I'm not sure this is really required. Can't we just ditch the path return or do you want it?

I was half/half on it.  Probably more simple to not include `wb_write()` and instead use `wb_save(wb)$path` or `wb |> wb_save() |> wb_get_path()` -- easy enough to add the getting function.

I'll make those updates.

R/class-workbook-wrappers.R
```diff
@@ -61,9 +65,14 @@ wb_workbook <- function(
 #' }
 wb_save <- function(wb, path = NULL, overwrite = TRUE) {
   assert_workbook(wb)
-  wb$clone()$save(path = path, overwrite = overwrite)$path
+  wb$clone()$save(path = path, overwrite = overwrite)
 }
 
+#' @rdname wb_save
+#' @export
```
jmbarbone, line 72: `wb_write()` is a bit cluttering.

Can remove and suggest: `wb_save(wb)$path` or include `wb_get_path()` (needs `wbWorkbook$get_path()`.

JanMarvin (approved):
LGTM!

Issue #324, opened 2022-09-12
fillMergedCells earlier in wb_to_df


JanMarvin:
if start column is called with fill merged cells, filling can fail with "replacement has 0 elements ..." error. This could be prevented, if the filling is called earlier in the function.

JanMarvin:
Have to come up with an example. Seen it with SAS XLSX output from proc tabulate.

I assume the issue was that I wanted the last row of a merged column as variable name and wanted to start reading from that row.

The issue might have been that I hadn't imported the upper cells, so wb_to_df was unable to convert the content.

JanMarvin:
Yet I'm not sure how to circumvent this. The solution might be an internal second wb_to_df call that reads the merged cells only and fills them. So that we always have valid cell content to insert and do not end up with unavailable content.

Though prior to solving this, I'd like to see my initial issue reproduced.

PR #441, opened 2022-11-26
Fix fill merged cells

sjewo:
Hi @JanMarvin,

i found a minor problem while reading merged cells. If the merged area only contains NA values the import fails with an error. I saw this error in combination with startRow (after selecting a different startRow the merged cells became all NA)

```
library(openxlsx2)
library(testthat)

wb <- wb_workbook()
wb$add_worksheet("sheet1")
wb$add_data(1, t(matrix(c(1:3, NA_real_), 4, 4)), startRow = 3, startCol = 1, colNames = FALSE)

wb$merge_cells(1, rows = 1:4, cols = 4)

tmp_file <- temp_xlsx()
wb_save(wb, tmp_file)

# runs fine
read_xlsx(tmp_file, fillMergedCells = FALSE)

# Error in x[[jj]][iseq] <- vjj : replacement has length zero
read_xlsx(tmp_file, fillMergedCells = TRUE)
```

A simple solution is checking if the merged cells are all NA and skip this cell range. I created a pull request with a possible solution and a new test.

### DIFF
--- a/R/wb_functions.R
+++ b/R/wb_functions.R
@@ -616,6 +616,11 @@ wb_to_df <- function(
       for (i in seq_along(mc)) {
         dms <- dims_to_dataframe(mc[i])
 
+        # Skip if merged cell is empty
+        if(all(is.na(z[rownames(z) %in% rownames(dms),
+                       colnames(z) %in% colnames(dms)])))
+          next
+
         z[rownames(z) %in% rownames(dms),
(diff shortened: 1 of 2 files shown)


JanMarvin:
Thanks, this might be related to #324. I never got to recreate the issue myself. Assumed that it was caused by merged cells overlapping into the requested region. Like every cell in row 1 is filled and a merged cell with column 2. Column 2 is empty, but the user requests to skip column 1 and `fillMergedCells = TRUE`.

Anyhow, nice catch and thanks for the PR :)

PR #449, opened 2022-11-29
provide add_drawing() for {rvg}

JanMarvin:
Example:

``` r
library(openxlsx2)
library(ggplot2)
library(rvg)

tmp <- tempfile(fileext = ".xml")

## rvg example
dml_xlsx( file =  tmp, fonts = list(sans = "Bradley Hand"))
ggplot(data = iris,
       mapping = aes(x = Sepal.Length, y = Petal.Width)) +
  geom_point() + labs(title = "With font Bradley Hand") +
  theme_minimal(base_family = "sans", base_size = 18)
dev.off()

wb <- wb_workbook()$
  add_worksheet()$
  add_drawing(xml = tmp)$
  add_drawing(xml = tmp, dims = "A9:H16")$
  add_drawing(xml = tmp, dims = NULL)
```

* [x] needs a wrapper function
* [x] needs tests 
* [x] provide one cell anchor option


JanMarvin:
Some parts of the drawing are out of the cell, but that seems to be related to the {rvg} code. We simply take whatever input we get and shove it into our document.

JanMarvin:
In the future we could provide an option in `wb_add_plot()` to create xml plots and embed the `rvg` call

PR #461, opened 2022-12-09
add calc_chain argument to wb_load. closes #432 

### LINKED ISSUE #432: calcChain improvements
ideas
* [ ] from an old comment in openxlsx: When deleting cells, `Reduce calcChain i attributes & remove calcs on sheet`
* [ ] add our formula cells to calcChain as well 
* [ ] allow writing a file without a calcChain. since libre office is not yet handling the full calc on load argument correctly, we could try to experiment with disabling the calc properties completely

JanMarvin:
Previously I wanted to improve calculation chain modifications in #432, but they remain a nuisance. Therefore this PR adds a new argument `calc_chain` to `wb_load()` to entirely ignore loading the calculation chain, which will be the default.

Basically
* the sheet id in the calculation chain must not match our internal sheet id. Adding or removing sheets on our end might change the id. Therefore if the user overwrites a cell in the calculation chain, we might update the wrong calculation chain cell. Ergo: unless someone puts in some work to really improve this situation, our way of dealing with the file remains error prone and will leave the user with the usual useless Excel error message.
* The calculation chain is only partially supported in other spreadsheet software. LibreOffice supports only a fraction, Numbers simply ignores it.

@nicojx This PR should fix the issue you mailed me about.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # openxlsx2 (in development)
 
+## New features
+
+* Provide new argument `calc_chain` to `wb_load()`. This is set to `FALSE` per default, to ignore the calculation chain if it is found. This change only reflects files imported from third party spreadsheet software and should not be visible to the user. [461](https://github.com/JanMarvin/openxlsx2/pull/461)
+
 ## Fixes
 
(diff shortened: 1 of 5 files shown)


JanMarvin:
Well that's the other issue #467 not the calcChain one. One per pull request

nicojx:
Ah! I'll try the other one then.

JanMarvin:
Hmm, not for me, and there is no more calc chain to complain about. The default is now to remove it completely from the file. It will be re-created by spreadsheet software once our output is loaded. This is the default approach openxlsx uses.
From your example above, the ampersand issue remains. If that cell is replaced with just character content it works fine for me.

nicojx:
Ah! You're right, it be good if I didn't mix my tests together! Thanks again... :)

PR #466, opened 2022-12-09
create chartsheets. closes 464

JanMarvin:
```R
library(openxlsx2)
library(mschart)

# broken unless mschart is added
wb <- wb_workbook()$
  add_worksheet("A & B")$
  add_data(x = mtcars)$
  add_chartsheet()

# create wb_data object
dat <- wb_data(wb, 1, dims = "A1:E6")

# call ms_scatterplot
data_plot <- ms_scatterchart(
  data = dat,
  x = "mpg",
  y = c("disp", "hp"),
  labels = c("disp", "hp")
)

wb$add_mschart(graph = data_plot)

wb$open()
```

~~* [ ] warning if no chart found on Chartsheet. sometime in the future we can provide a default Chartsheet and update the Chartsheet if a chart is assigned to an already existing Chartsheet~~
* [ ] update NEWs


JanMarvin:
@jmbarbone could you have a look why, in the example above, the sheet has to be selected? If I write this `wb$add_mschart(graph = data_plot)`, `current_sheet()` evaluates to `1` and the workbook is damaged. It must be caused by the waiver code, but I could not understand what was going on.
If changing the behavior would cause other troubles, we can keep things the way they are. Most likely adding chartsheets is not of broad user interest.

jmbarbone:
@JanMarvin The current sheet is reset in `wb_data()` because we call something there.  This I guess is more like a "last_called_sheet()" which doesn't seem right.

``` r
library(openxlsx2)
wb <- wb_workbook()
wb$.__enclos_env__$private$current_sheet
#> [1] 0
wb$add_worksheet()
wb$.__enclos_env__$private$current_sheet
#> [1] 1
wb$add_data(x = mtcars)
wb$.__enclos_env__$private$current_sheet
#> [1] 1
wb$add_chartsheet()
wb$.__enclos_env__$private$current_sheet
#> [1] 2

# create wb_data object
dat <- wb_data(wb, 1, dims = "A1:E6")
wb$.__enclos_env__$private$current_sheet
#> [1] 1
```

<sup>Created on 2022-12-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Might also be worth it to have a `$get_current_sheet()` method to check this a little more easily

JanMarvin:
Ah thanks, I completely missed that.

PR #468, opened 2022-12-09
Fix writing xml escaped characters. fixes #467

### LINKED ISSUE #467: Writing & (ampersand) causes error
Writing a value with an ampersand (&) causes an error when opening the workbook:

```R
wb_workbook()$
  add_worksheet()$
  add_data(dims = "A1", x = "A & B")$
  open()
```

The 'repair' error message says: 
`Replaced Part: /xl/worksheets/sheet1.xml part with XML error.  Illegal name character. Line 1, column 1063.`

JanMarvin:
This reverts the previous "fix"

### DIFF
--- a/R/wb_functions.R
+++ b/R/wb_functions.R
@@ -521,13 +521,13 @@ wb_to_df <- function(
   # text in t
   if (any(cc_tab %in% c("inlineStr"))) {
     sel <- cc$c_t %in% c("inlineStr")
-    cc$val[sel] <- replaceXMLEntities(is_to_txt(cc$is[sel]))
+    cc$val[sel] <- is_to_txt(cc$is[sel])
     cc$typ[sel] <- "s"
   }
   # test is sst
   if (any(cc_tab %in% c("s"))) {
     sel <- cc$c_t %in% c("s")
-    cc$val[sel] <- replaceXMLEntities(sst[as.numeric(cc$v[sel]) + 1])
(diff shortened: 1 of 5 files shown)


tests/testthat/test-strings_xml.R
```diff
@@ -89,4 +89,15 @@ test_that("strings_xml", {
...
+
+  amp <- temp_xlsx()
+  wb_workbook()$
+    add_worksheet()$
+    add_data(dims = "A1", x = "A & B")$
+    save(amp)
+
+  exp <- "A & B"
+  got <- wb_to_df(amp, colNames = FALSE)[1, 1]
+  expect_equal(exp, got)
```
JanMarvin, line 101: maybe I should check the XML input. this is the test from the previous PR. Any we know how well that went

R/wb_functions.R
```diff
@@ -521,13 +521,13 @@ wb_to_df <- function(
   # text in t
   if (any(cc_tab %in% c("inlineStr"))) {
     sel <- cc$c_t %in% c("inlineStr")
-    cc$val[sel] <- is_to_txt(cc$is[sel])
+    cc$val[sel] <- replaceXMLEntities(is_to_txt(cc$is[sel]))
```
JanMarvin, line 524: I'm not entirely happy with this. It should work reasonably well, but maybe we can check it we have to replace something in the first place.

PR #474, opened 2022-12-13
473 na strings

### LINKED ISSUE #473: consider `na_string()` default for `na.strings`
waiver function for explicit `na.strings` setting

```r
na_string <- function() {
  structure("na_string", class = "openxlsx2_waiver")
}

is_na_string <- function(x) {
  is_waiver(x) && isTRUE(x == "na_string")
}
```

@JanMarvin , I did some testing and was able to implement this without having to update any of the tests.  I'm pretty sure my poor handling of these is what's wrong with #471 . 

This would help with using explicit values (definitely my preference; and we only have to check this _once_) but also (with another small internal adjustment) allow users to set their default strings value in `options()`. 

If we would want to use

``` r
foo <- function(
    ...
    na.strings = getOption("openxlsx2.na.strings", na_strings())
    ...
) {
  ...
}
```

We could allow zero length vectors (`character())` to replace the `NULL` behavior within `write_data2()`; which currently errors through `txt_to_is(character())`.  Explicitly setting `na.strings = NULL` would have the same effect; as would not setting it.

jmbarbone:
resolves #473

A much cleaner PR.  Nothing in the tests has to be changed for this to work.

Documentation would need to be updated if we wanted to go through with this.  Happy to wait until a later release.

- add na_strings() and is_na_strings()
- implement na_strings()
- redoc

### DIFF
--- a/R/class-workbook-utils.R
+++ b/R/class-workbook-utils.R
@@ -158,6 +158,15 @@ next_sheet <- function() {
   structure("next_sheet", class = "openxlsx2_waiver")
 }
 
+na_strings <- function() {
+  structure("na_strings", class = "openxlsx2_waiver")
+}
+
 is_waiver <- function(x) {
   inherits(x, "openxlsx2_waiver")
 }
+
+is_na_strings <- function(x) {
+  is_waiver(x) && isTRUE(x == "na_strings")
+}
+
--- a/R/class-workbook-wrappers.R
+++ b/R/class-workbook-wrappers.R
@@ -114,12 +114,9 @@ wb_add_data <- function(
     sep             = ", ",
     applyCellStyle  = TRUE,
     removeCellStyle = FALSE,
-    na.strings
+    na.strings = na_strings()
 ) {
   assert_workbook(wb)
-
-  if (missing(na.strings)) na.strings <- substitute()
-
   wb$clone(deep = TRUE)$add_data(
     sheet           = sheet,
     x               = x,
@@ -201,11 +198,9 @@ wb_add_data_table <- function(
     bandedCols  = FALSE,
     applyCellStyle  = TRUE,
     removeCellStyle = FALSE,
-    na.strings
+    na.strings = na_strings()
 ) {
   assert_workbook(wb)
-  if (missing(na.strings)) na.strings <- substitute()
-
   wb$clone()$add_data_table(
     sheet       = sheet,
     x           = x,
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -982,10 +982,8 @@ wbWorkbook <- R6::R6Class(
         sep             = ", ",
         applyCellStyle  = TRUE,
         removeCellStyle = FALSE,
-        na.strings
+        na.strings = na_strings()
       ) {
-
-      if (missing(na.strings)) na.strings <- substitute()
 
       write_data(
         wb              = self,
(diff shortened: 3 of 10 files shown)


jmbarbone:
> Sorry, at my first coffee this morning, will this modify the behavior aside from replacing missing? Aka if I want to write "N/A", do I have to write `na_strings("N/A")`?
> 
> Or is this just something like our flavor of `rlang::missing_arg()`?

More `rlang::missing_arg()` flavor.  Would it make sense to allow something like `na_strings("N/A")` or `na_strings("MY_NA_VALUE")` to write into the cell?  Default would then be `na_strings("#N/A")` and that "`#N/A"` value is directly placed into the cell.

https://github.com/JanMarvin/openxlsx2/blob/2a8dbaff89b9e5f1695e2cb433ae415fe6aa3499/R/write.R#L311-L330

JanMarvin:
Thanks, it was just for my understanding. I'll review it after the release. Should be harmless and I prefer it the way it is,  but I don't have the time and focus right now. One thing at the time :)

R/class-workbook-utils.R
```diff
@@ -158,6 +158,15 @@ next_sheet <- function() {
   structure("next_sheet", class = "openxlsx2_waiver")
 }
 
+na_strings <- function() {
```
JanMarvin, line 161: Since this is visible to the user, it probably should have a corresponding man entry?

JanMarvin (changes requested):
* Since `na_strings()` will be visible to the user, we should have a manual entry.
* Minor `lintr` issues
* Maybe update NEWS for the argument change?
* Add this as a default for `wb_to_df()` as well? So that the user has a clean interface? Something like `wb_to_df(..., na.strings = na_strings()) { ..., if (is_na_strings) na.strings <- "#N/A")`.

R/class-workbook-utils.R
jmbarbone: It might make sense to export the other waivers, too.  Currently `current_sheet()` and `next_sheet()` are also internals only.

JanMarvin: Yes, good idea. Unfortunately documentation is not our strongest side :)

JanMarvin (approved):
Looks fine to me. What about my previous comment:

>     * Add this as a default for `wb_to_df()` as well? So that the user has a clean interface? Something like `wb_to_df(..., na.strings = na_strings()) { ..., if (is_na_strings) na.strings <- "#N/A")`.

JanMarvin:
@jmbarbone Please let me know what you intend to do with these patches. I like this `na_strings()` approach (I was actually thinking that something like a more general way like `rlang::missing_arg()` might even be preferable), but I personally can live without both.
So if you want it, I'll help where I can, but you'll have to update and prepare both. If you are no longer interested, I would close this and #471.

jmbarbone:
@JanMarvin , sorry, lost track of this.  I've made a few more edits and it looks like everything is running fine after merging in `main`.

JanMarvin:
Thanks, I'll have a look tomorrow.

JanMarvin:
Thank you, I have merged this PR. Looks nice :)

But please, before you start updating #471, I still haven't made up my mind about the size and impact of the PR and would veto the merge at this point. I still feel that the risk of breaking things along the way still outweighs the benefit of having more organized (and possibly cleaner) code. Since I started porting some real life work over to `openxlsx2` (which exposed some of the bugs I have been fixing lately), I don't like having a nice but broken mess and prefer a working mess, if you know what I mean.

Maybe we can break #471 into smaller pieces (e.g. not including the unnecessary `testthat` changes) or go function by function, but just having one big pull request that changes 61 files and moves ~5k lines of code is not what I like at the moment. And even if it does not make things better, I am really sorry to be so picky about this, especially since you have already spent some time on it.

PR #497, opened 2022-12-24
fix reading of xml escapes

JanMarvin:
This is https://github.com/ycphs/openxlsx/issues/394 (found it via https://asterhu.com/import-excel-formula-with-ampersand-to-r/)

Could you check if this solves your issue @aster-hu?

* [x] add test 
* [x] compare speed with internal stringi solution 
* [x] use the faster one in other places as well? if not, why? to much hassle or generally safer?
* [ ] ~~argument, `is_xml` to skip XML creation ?~~


aster-hu:
It resolves my issue, thanks a lot!

PR #499, opened 2022-12-26
Write sst with `inline_string = FALSE`

### LINKED ISSUE #67: convert characters to sharedStrings when saving
Not sure yet.

Excel saves inlineStr as sharedStrings per default. most likely because of memory reasons. We could provide some interface for saving sharedStrings too. Similar to what `openxlsx` did.

1. writeData should convert characters to sharedStrings
2. sharedStrings needs to be searched for the last id or matching entries
3. sst and our sharedStrings should be merged
4. the IDs must be updated
5. maybe provide some way to modify the sst entry for style modifications
6. is it worth the hassle?

JanMarvin:
closes #67 

## what is this about?

Currently we write every character as `inlineStr`. Therefore we have a couple of cells like this
 ```XML
<c t="inlineStr">
  <is>
    <t>foo</t>
  </is>
</c>
```

This is completely fine with the open xml standard. But Excel usually writes characters as `sharedStrings`. They provide a `sharedStrings.xml` file looking like this:

```XML
<sst uniqueCount="1">
  <si>
    <t>foo</t>
  </si>
</sst>
```

and a cell that looks like this:

```XML
<c t="s">
  <v>0</v>
</c>
```

## update

I have decided to go a different route. For now we treat `sharedStrings` similar to `inlineStrings` and simply create the shared strings once we are done with adding data.

* write strings as `<is><t>...</t></is>`  to `cc`
* check for unique strings in `cc`
* combine these with `wb$sharedStrings`
* replace `cc` values with index positions
* update `cc` and `wb$shareStrings`

This brings the benefit of having shared strings over the workbook which can reduce the overall file size with string duplicates in a workbook. But, since we do not check for these initially, it is a bit slower, because of the `unique` and `match` calls once writing is done. On the other side it brings the benefit of potentially smaller files and the code is a lot cleaner.

``` r
library(openxlsx2)

res <- microbenchmark::microbenchmark(
  wb_workbook()$add_worksheet()$add_data(x = iris, inline_strings = FALSE),
  wb_workbook()$add_worksheet()$add_data(x = iris, inline_strings = TRUE),
  times = 100
); res
#> Unit: milliseconds
#>                                                                      expr
#>  wb_workbook()$add_worksheet()$add_data(x = iris, inline_strings = FALSE)
#>   wb_workbook()$add_worksheet()$add_data(x = iris, inline_strings = TRUE)
#>       min       lq     mean   median       uq       max neval
#>  5.419856 5.581581 6.763844 5.901882 6.227162 54.860594   100
#>  5.115616 5.287041 5.785157 5.533161 5.982857  9.783031   100
```


R/write.R
```diff
@@ -292,8 +294,52 @@ write_data2 <- function(
...
+      data[sel] <- lapply(data[sel], function(x) as.character(match(x, unis) + base_len - 1L))
+      if (colNames) {
+        sel <- as.character(data[1, ]) %in% unis
+        data[1, sel] <- lapply(data[1, sel], function(x) as.character(match(x, unis) + base_len - 1L))
+      }
+
+      new_sst <- vapply(unis, txt_to_si, NA_character_, USE.NAMES = FALSE)
+      wb$sharedStrings <- c(as.character(wb$sharedStrings), new_sst)
+      attr(wb$sharedStrings, "uniqueCount") <- length(unis)
+      attr(wb$sharedStrings, "text") <- unis
```
JanMarvin, line 334: needs to be extended from previous text.

Issue #226, opened 2022-06-09
High-Level style helpers


JanMarvin:
* [x] `add_border()` #220 (provides only the options I considered helpful)
* [x] `add_fill()` #222
* [x] `add_font()` #228
* [x] `add_numfmt()` #229
* [x] `add_cell_style()` #230 various options, similar to `create_cell_style()`. might replace bare metal functions in previous functions. Might look into cellstyle xf
* [ ] `add_sheet_style()` ( ... maybe?)

This reduces the first example shown here: https://janmarvin.github.io/openxlsx2/articles/openxlsx2_style_manual.html

```R
# all functions need a canvas to draw on
wb <- wb_workbook()$add_worksheet("S1", gridLines = FALSE)$add_data("S1", mtcars)
# add a thick border below the colnames
wb$add_border("S1", dims = "A1:K1", bottom_border = "thick", bottom_color = c(theme = "4"))
# fill the back of the colnames
wb$add_fill("S1", dims = "A1:K1", color = c(theme = "3"))
# change the font of the colnames
wb$add_font("S1", dims = "A1:K1", color = c(theme = "2"), name = "Arial", size = "16")
# add number format
wb$add_numfmt("S1", dims = "A1:A33", numfmt = 2)
# add another number format
wb$add_numfmt("S1", dims = "F1:F33", numfmt = "#.0")
# open the workbook
wb$open()
```

JanMarvin:
Hmm, I plan to work on the remaining functions tomorrow (`add_numfmt()` and a simple `add_cell_style()` for centering, for example). Am a bit tired today, it's been a long week already. After that, we can always talk about improvements. Reducing code is more than welcome and I'd be very happy if you go through this code mess once I'm done. You are a bit better at creating user friendly code :) (and someone needs to add the `wb_wrapper` tests, hehe).

However, I'm not sure why you want this specs code. Maybe I'm missing something, but I don't see any advantages in the design you created. I still want to revive some of the options (afaik there were or are some for frame colors), but otherwise I'm not sure what else these specs would bring.

jmbarbone:
> You are a bit better at creating user friendly code :)

Probably just really picky and good at over-engineering.

> I'm not sure why you want this specs code. 

A few reasons I can think of:
* Move entry validation into separate functions -- moving it outside of the R6 method might make it easier to review -- easier to trace the error codes to the spec building and not distract from other parts of the function.  
[...]

PR #501, opened 2022-12-28
Colour color pt1

JanMarvin:
A first part to fix the `color`/`colour` mess of ours.

* We currently have functions using arguments `color` or `colour`
* We have `wb_colour()`
* R uses both, ggplot and many others use both too

In a first attempt to align the naming in `openxlsx2` I have added `...` argument to (hopefully) all functions using `color`/`colour` arguments. Every function calling one of these should trigger a `standardise_color_names()` or `standardize_colour_names()` function. They convert every `colour` into a `color` argument and vice versa.

In a second part we should switch our default to `color`. After all this is used by open xml under the hood. Though that will be a larger and messy search & replace PR.

```R
  wb_color <- wb_workbook() %>%
    wb_add_worksheet(tabColor = "green") %>%
    wb_add_fill(color = wb_color("blue")) %>%
    wb_add_border(
      dims   = "G12:H13",
      left_color   = wb_color("red"),
      right_color  = wb_color("blue"),
      top_color    = wb_color("green"),
      bottom_color = wb_color("yellow")
    )

  wb_colour <- wb_workbook() %>%
    wb_add_worksheet(tabColour = "green") %>%
    wb_add_fill(colour = wb_colour("blue")) %>%
    wb_add_border(
      dims   = "G12:H13",
      left_colour   = wb_colour("red"),
      right_colour  = wb_colour("blue"),
      top_colour    = wb_colour("green"),
      bottom_colour = wb_colour("yellow")
    )
```


R/class-color.R
```diff
@@ -35,4 +34,51 @@ wb_colour <- function(
...
+wb_color <- wb_colour
+
 is_wbColour <- function(x) inherits(x, "wbColour")
+
+#' takes color and returns colour
+#' @param ... ...
+#' @returns named colour argument
+#' @keywords internal
+#' @noRd
+standardise_color_names <- function(...) {
```
jmbarbone, line 49: Do I understand this correctly?

```r
foo <- function(color, ...) {
  standardized_color_name(...)  
  color
}

foo(color = 1)
#> [1] 1

foo(colour = 1)
#> [1] 1
```

Not sure how much I like messing around with re-assigning values.  A simpler implementation could be:

``` r
foo <- function(color = colour, colour) {
  check_color_colour(color, colour)
  color
}

check_color_colour <- function(color, colour) {
  if (missing(colour)) {
    return(invisible())
  }
  
  if (!identical(color, colour)) {
    stop("both `color` and `colour` cannot be set")
  }
}

# setting only color
foo(1)
#> [1] 1

# setting only colour (turns into color)
foo(colour = 1)
#> [1] 1

# both are identical, that's fine
foo(color = 1, colour = 1)
#> [1] 1

# difference, we have a problem
try(foo(color = 1, colour = 2)) 
#> Error in check_color_colour(color, colour) : 
#>   both `color` and `colour` cannot be set
```

<sup>Created on 2022-12-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>
We'd have to include `@params color, colour` but it would keep it explicitly clear to the user.

JanMarvin: 1) Yes, that is all it does.
2) I don't like duplicating all arguments. It creates visual confusion, especially with the border stuff and all the different colors. We'll end up with 10 different color arguments and a bunch of comparison functions. Therefore I mimicked the `ggplot2::aes()` approach. The difference to their approach is that they only have a single argument to care about, while we have various. But true, I haven't considered the `foo(color = 1, colour = 2)` approach. Might want to throw a warning if this exists in `parent.frame()`

JanMarvin: Update: 12 different colors ... we have two for the inner grid and checking with `exists()` is not enough,  because they all exist at this stage. Maybe this will change once you get to #226 *\*winkwink\**.
Therefore I consider keeping it the way it is. After all no harm is done this way. If the user ends up with a unexpected color ... there are more important things to worry about. If their input creates no color at all, fine too. I'd love for R to have a dynamic way to handle this case. Otherwise I simply strive for a consistent user interface.

JanMarvin:
Tomorrow I intend to merge #501 and #502 with main. Could be a nightmare to merge with #471, but there are still a couple of winter evenings left

PR #508, opened 2022-12-30
fix diagonal in create border

JanMarvin:
We can have one line style and either up, down or both.

```R
  add_border(dims = "A1", diagonal = "medium", diagonal_type = c("down"))$
  add_border(dims = "A2", diagonal = "medium", diagonal_type = c("up"))$
  add_border(dims = "A3", diagonal = "medium", diagonal_type = c("down", "up"))$
  add_border(dims = "A4")
```

Probably `diagonal`, `diagonal_color` and `diagonal_type = c("up", "down")` would be a better way to approach this.


JanMarvin:
should be simpler once add_border uses `...`

PR #516, opened 2023-01-17
[wb_add_data] string_nums option. closes #503

### LINKED ISSUE #503: Write number as character without Excel error
In the example below we write `1992` three times: (i) as character, (ii) as numeric with a style format and (iii) as numeric. This is the output from Excel:

Screenshot 2022-12-28 at 18 44 47

The first cell contains a number written as character, therefore the well known Excel green flag error. The second is the Excel way of writing a number as character. This uses a special style designed for this. Finally, the third is simply a number.

We might want to look into ways to write the correct way per default. Tough currently I am not sure how to approach this efficiently.

```R
library(openxlsx2)


  add_data(dims = "A1", x = "1992")$
  add_data(dims = "A2", x = 1992)$
  add_data(dims = "A3", x = 1992)

wb$add_cell_style(dims = "A2", applyNumberFormat = "1", quotePrefix = "1", numFmtId = "49")

# wb$open()
```

JanMarvin:
This PR provides a `string_nums` option for `openxlsx2` to control if strings that can be expressed as numerics should be written as styled numerics and not as character strings (e.g. `"1"` as `1`). ~~This does not provide on the fly conversions from strings to numerics (which ... would be possible as well; simply avoid writing the style to the cell).~~

This needs some testing (accuracy and speed) and for now I have hidden it with an option. This option knows states 0 (the current default), 1 (write string as number with cell style) and 2 (write string as number). The entire conversion part likely will have an impact on writing time, due to the call to `is_double()`.

``` r
library(openxlsx2)

dat <- data.frame(x = "2023", y = 2023)
wb <- wb_workbook()

options("openxlsx2.string_nums" = 1)

wb <- wb %>% 
  wb_add_worksheet() %>% 
  wb_add_data(x = dat)

options("openxlsx2.string_nums" = 0)

wb <-  wb %>% 
  wb_add_worksheet() %>% 
  wb_add_data(x = dat)
  

# new
str(wb_to_df(wb, 1))
#> 'data.frame':    1 obs. of  2 variables:
#>  $ x: num 2023
#>  $ y: num 2023
#>  - attr(*, "tt")='data.frame':   1 obs. of  2 variables:
#>   ..$ x: chr "n"
#>   ..$ y: chr "n"
#>  - attr(*, "types")= Named num [1:2] 1 1
#>   ..- attr(*, "names")= chr [1:2] "A" "B"

# old
str(wb_to_df(wb, 2))
#> 'data.frame':    1 obs. of  2 variables:
#>  $ x: chr "2023"
#>  $ y: num 2023
#>  - attr(*, "tt")='data.frame':   1 obs. of  2 variables:
#>   ..$ x: chr "s"
#>   ..$ y: chr "n"
#>  - attr(*, "types")= Named num [1:2] 0 1
#>   ..- attr(*, "names")= chr [1:2] "A" "B"
```

## Update
With the latest push on my Arch Linux desktop I see the following. It is quite fast, though I'm not sure about the effects of setting this style `applyNumberFormat = "1", quotePrefix = "1", numFmtId = "49"` on every character cell, if any cell requires  it. That somehow seems like overkill. Setting it only to those cells that need it, slows the example terribly down.

``` r
n <- 10000
k <- 10

mm <- matrix(rnorm(n = n * k), nrow = n, ncol = k)
mc <- matrix(as.character(mm), nrow = n, ncol = k)


library(openxlsx2)

wb <- wb_workbook()

t1 <- Sys.time()
options("openxlsx2.string_nums" = 0)
wb$add_worksheet()$add_data(x = mc)
t2 <- Sys.time()
options("openxlsx2.string_nums" = 1)
wb$add_worksheet()$add_data(x = mc)
t3 <- Sys.time()
options("openxlsx2.string_nums" = 2)
wb$add_worksheet()$add_data(x = mc)
t4 <- Sys.time()

t2 - t1 # initial matrix as character
#> Time difference of 0.6670108 secs
t3 - t2 # matrix as numeric with style
#> Time difference of 0.963567 secs
t4 - t3 # matrix as numeric
#> Time difference of 0.3893292 secs
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,13 @@
 
 * Add `wb_copy_cells()` a wrapper that allows copying cell ranges in a workbook as direct copy, as reference or as value. [515](https://github.com/JanMarvin/openxlsx2/pull/515)
 
+* Experimental option: `openxlsx2.string_nums` to write string numerics differently. A string numeric is a numeric in a string like: `as.character(1.5)`. The option can be
+  * 0 = the current default. Writes string numeric as string (the incorrect way according to spreadsheet software)
+  * 1 = writes string numeric as numeric with a character flag (the correct way according to spreadsheet software)
+  * 2 = convert all string numeric to numeric when writing
+  
+  This is experimental, because the impact is somewhat unknown. It might trigger unintended side effects. Feedback is requested.
+
 ## Fixes
 
 * Reading of files with frozen panes and more than one section node was restored. [495](https://github.com/JanMarvin/openxlsx2/pull/495)
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -29,8 +29,8 @@ long_to_wide <- function(z, tt, zz) {
     invisible(.Call(`_openxlsx2_long_to_wide`, z, tt, zz))
 }
 
-wide_to_long <- function(z, vtyps, zz, ColNames, start_col, start_row, ref) {
-    invisible(.Call(`_openxlsx2_wide_to_long`, z, vtyps, zz, ColNames, start_col, start_row, ref))
+wide_to_long <- function(z, vtyps, zz, ColNames, start_col, start_row, ref, string_nums) {
+    invisible(.Call(`_openxlsx2_wide_to_long`, z, vtyps, zz, ColNames, start_col, start_row, ref, string_nums))
 }
 
 #' @param colnames a vector of the names of the data frame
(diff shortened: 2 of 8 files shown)


JanMarvin:
hmpf. the Mac build is falling due to the sprintf warnings triggered by clang. They are already addressed in the Rcpp repo and fortunately simply annoying and should be fixed once the next Rcpp release hits CRAN.

PR #523, opened 2023-01-27
Speedup add data pt2

JanMarvin:
pt2 this should be checked with old non utf8 windows


JanMarvin:
Closing, this did not work. Found by the windows test :)

PR #524, opened 2023-01-27
Speedup add data pt3

JanMarvin:
further speedups.

This replaces the code used in `wide_to_long()` with the one used in `loadvals()`. This is generally quicker, but it is quicker due to not caring about the write order of the dataset. Basically this creates a vector of a vector of strings and returns a data frame. Each vector of strings is a column. All the vectors are the data frame.
This is faster than the previous approach where we create a data frame and assign the values of the data set in the correct order.

Replacing the original approach currently causes errors in the tests because the order is different than what the tests expect.

Overall, I have seen speed improvements of ~50% on larger data sets. However, our approach is not suitable for large data sets. Excel can handle n = 1,048,576 rows and k = 16,384 columns. I couldn't write more than a `n x 50` matrix of `1`s to a workbook on a 16GB machine. So this works:

```R
mm <- matrix(1, nrow = 1048576, ncol = 50)

library(openxlsx2)

wb <- wb_workbook()$add_worksheet()$add_data(x = mm, colNames = FALSE, applyCellStyle = FALSE)
```

The issue is that we have the initial data frame and pivot it longer, to create our `cc` data frame. Once this data frame is created, we can write it to xml. However, at this point we already have two big data frames and this story will not have a happy ending ...

If you want to write large amounts of data, you should use a binary format and a writer that transfers the data directly to the drive, or use a database. Use anything but open xml files.

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -29,8 +29,8 @@ long_to_wide <- function(z, tt, zz) {
     invisible(.Call(`_openxlsx2_long_to_wide`, z, tt, zz))
 }
 
-wide_to_long <- function(z, vtyps, zz, ColNames, start_col, start_row, ref, string_nums, inline_strings) {
-    invisible(.Call(`_openxlsx2_wide_to_long`, z, vtyps, zz, ColNames, start_col, start_row, ref, string_nums, inline_strings))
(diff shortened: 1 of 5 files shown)


JanMarvin:
Closed in favor of pt4

PR #536, opened 2023-02-03
Cleanup drawings and vmlDrawings reference

JanMarvin:
Just the sort of pull requests I like, touches all sorts of places of code. So it's also cluttered to the max. Joking aside, there is now a new field relships that contains the relationships of a worksheet. This can be addressed and used to manage drawings and vml.

* [x] This needs testing in Excel
* [x] More tests
* [x] update NEWS


JanMarvin:
Merging this, because it improves the current situation, but there might still be corner stones I haven't thought about.

PR #555, opened 2023-02-28
[wb_to_df] modify skipEmptyCols/skipEmptyRows. closes #554

### LINKED ISSUE #554: [FR] skipEmptyColumn should keep empty named columns
It was requested that `openxlsx2` could keep empty columns with `wb_to_df(skipEmptyColumn = TRUE)` so that only columns which are `NA`s are kept. This should be doable, we only have to move the output creating part of `wb_to_df()` behind the `skipEmtpyColumn` part.

``` r
library(openxlsx2)

# feature request skipEmptyCols skips should keep named empty columns

# removes empty column Species
# dat <- wb_to_df(wb, skipEmptyCols = TRUE)

## initialize empty cells
na_mat <- matrix(NA, nrow = 22, ncol = 7)

## create artificial data with empty column
dat    <- iris[1:20,]
dat$Species <- NA

## create the workbook
wb <- wb_workbook()$
  add_worksheet()$
  add_data(x = na_mat, colNames = FALSE)$
  add_data(x = dat)


# workaround
header <- wb_to_df(wb, rows = 1, colNames = FALSE, skipEmptyCols = TRUE)
dat <- read_xlsx(wb, cols = names(header))

head(dat)
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#> 2          5.1         3.5          1.4         0.2      NA
#> 3          4.9         3.0          1.4         0.2      NA
#> 4          4.7         3.2          1.3         0.2      NA
#> 5          4.6         3.1          1.5         0.2      NA
#> 6          5.0         3.6          1.4         0.2      NA
#> 7          5.4         3.9          1.7         0.4      NA
```

JanMarvin:
The part was moved up in the function and is now before `rowNames` and `colNames`. Therefore it allows reading empty named columns and remove only columns that have no header.

This is a breaking change, because it changes the default behavior of the output. I have nothing against this, but I am not really a user of the `skipEmptyCols`/`skipEmptyRows` argument. Do you have any suggestions @jmbarbone ?

### DIFF
--- a/R/wb_functions.R
+++ b/R/wb_functions.R
@@ -595,10 +595,6 @@ wb_to_df <- function(
   if (any(zz$val == "", na.rm = TRUE)) zz <- zz[zz$val != "", ]
   long_to_wide(z, tt, zz)
 
-  # prepare colnames object
-  xlsx_cols_names <- colnames(z)
-  names(xlsx_cols_names) <- xlsx_cols_names
-
   # backward compatible option. get the mergedCells dimension and fill it with
   # the value of the first cell in the range. do the same for tt.
   if (fillMergedCells) {
@@ -638,6 +634,30 @@ wb_to_df <- function(
     }
 
   }
+
+  # is.na needs convert
+  if (skipEmptyRows) {
+    empty <- apply(z, 1, function(x) all(is.na(x)), simplify = TRUE)
+
+    z  <- z[!empty, , drop = FALSE]
+    tt <- tt[!empty, , drop = FALSE]
+  }
+
+  if (skipEmptyCols) {
+
+    empty <- vapply(z, function(x) all(is.na(x)), NA)
+
+    if (any(empty)) {
+      sel <- which(names(empty) %in% names(empty[empty == TRUE]))
+      z[sel]  <- NULL
+      tt[sel] <- NULL
+    }
+
+  }
+
+  # prepare colnames object
+  xlsx_cols_names <- colnames(z)
+  names(xlsx_cols_names) <- xlsx_cols_names
 
   # if colNames, then change tt too
   if (colNames) {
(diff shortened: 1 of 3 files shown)


tests/testthat/test-read_from_created_wb.R
```diff
@@ -131,3 +131,26 @@ test_that("reading with multiple sections in freezePane works", {
   wb$save(temp)
   expect_silent(wb <- wb_load(temp))
 })
+
+test_that("skipEmtpyCols keeps emty named columns", {
```
JanMarvin, line 135: empty

R/wb_functions.R
```diff
@@ -639,6 +635,30 @@ wb_to_df <- function(
 
   }
 
+  # is.na needs convert
+  if (skipEmptyRows) {
+    empty <- apply(z, 1, function(x) all(is.na(x)), simplify = TRUE)
```
jmbarbone, line 640: `simplify` was added to `apply()` in **R 4.1.0**; current dependency is `R (>= 3.4.0)`.

JanMarvin: Oh woops, thanks. It's in there for a long long time.
Otherwise any gut feeling whether or not we should fulfill this feature request?

R/wb_functions.R
```diff
@@ -639,6 +635,30 @@ wb_to_df <- function(
...
+    z  <- z[!empty, , drop = FALSE]
+    tt <- tt[!empty, , drop = FALSE]
+  }
+
+  if (skipEmptyCols) {
+
+    empty <- vapply(z, function(x) all(is.na(x)), NA)
+
+    if (any(empty)) {
+      sel <- which(names(empty) %in% names(empty[empty == TRUE]))
```
jmbarbone, line 651: I think this is the same as `which(empty)`?

jmbarbone (approved):

R/wb_functions.R
jmbarbone: Keeping named `NA` columns seems better than accidentally dropping them.  Users can always drop the columns afterwards.

R/wb_functions.R
JanMarvin: Ah yes thanks for spotting!

PR #569, opened 2023-03-29
avoid dot functions of R > 3.4.0. closes #568

### LINKED ISSUE #568: consider raising the version of R in Depends section of package DESCRIPTION
Dear openxlsx2 devs:

first of all thanks for a great package! It helps me immensely in my daily work.

Secondly: consider raising the required version of R in DESCRIPTION; it currently reads as R (>= 3.4.0)

This feels not quite correct, given that the package uses function `...names()` which only comes from R 4.1.0

https://github.com/JanMarvin/openxlsx2/blob/50a4248adf4343a76cc4696fc42bd623fd2e4808/R/class-color.R#L52

It may be a good idea to raise the required R version in some of the future releases (R 3.4.0 is quite ancient by now, and even 4.1.0 feels a bit dated).

Thanks for your consideration!

JanMarvin:
untested

### DIFF
--- a/R/class-color.R
+++ b/R/class-color.R
@@ -49,7 +49,7 @@ is_wbColour <- function(x) inherits(x, "wbColour")
 #' @noRd
 standardize_color_names <- function(...) {
 
-  # since R 4.1.0 got <- ...names()
+  # since R 4.1.0: ...names()
   got <- names(list(...))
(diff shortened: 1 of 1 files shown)


JanMarvin:
there appears to be an issue with actions. Well we are in no rush.

Issue #576, opened 2023-04-07
Allow add_image function to be a twoCellAnchor


rfineman:
Currently, the [add_image](https://github.com/JanMarvin/openxlsx2/blob/84eba654c80b7c366240812d7dcdb10e88cf46b9/R/class-workbook.R#L3710) function assumes a [oneCellAnchor](https://github.com/JanMarvin/openxlsx2/blob/84eba654c80b7c366240812d7dcdb10e88cf46b9/R/class-workbook.R#L3806). It would be nice to allow a twoCellAnchor where you specify both the start and end positions. This will circumvent any issues trying to resolve image width with column width.

PR #577, opened 2023-04-07
576 add twoCellAnchor option to add_image function

rfineman:
Resolves enhancement #576.

- update wb_add_image wrapper function
- update add_image internal function in workbook
- update add_plot internal function in workbook

### DIFF
--- a/R/class-workbook-wrappers.R
+++ b/R/class-workbook-wrappers.R
@@ -2118,10 +2118,15 @@ wb_set_last_modified_by <- function(wb, LastModifiedBy) {
 #' @param height Height of figure.
 #' @param startRow Row coordinate of upper left corner of the image
 #' @param startCol Column coordinate of upper left corner of the image
-#' @param rowOffset offset within cell (row)
-#' @param colOffset offset within cell (column)
+#' @param startRowOffset offset within cell (starting row)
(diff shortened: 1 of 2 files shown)


rfineman:
@JanMarvin if you like this enhancement, wasn't sure if you were comfortable renaming rowOffset/colOffset to startRowOffset/startColOffset as it would lead to breaking change. If preferred, we can keep rowOffset/colOffset and then also include startRowOffset/startColOffset that take rowOffset/colOffset as the default value.

Additionally, only changed add_plot to act the same way it currently operations. We could modify that as well, however have the additional hitch of how we wish to handle xy paramater.

JanMarvin:
Thanks for the issue and the pull request. I'll have a look tomorrow or after Easter. For drawings we have something in place that selects one or two cell anchor depending on dims. No dims is absolute, single cell is one cell anchor and cell range is two cell anchor. Can't we use something like that for images as well?

JanMarvin:
Hi @rfineman , thanks again for the PR. Based on your suggestion I have pushed #578 which supersedes this PR. Please let me know if this is what you are looking for.

rfineman:
@JanMarvin like your solution. works perfectly for my need. thanks.

PR #593, opened 2023-04-26
[wb_add_formula] updates to formulas including vignette. closes #490

### LINKED ISSUE #490: Improve add_formula()
Allow passing
* cm (those "hidden" array formulas)
* ref as in #489

JanMarvin:
add arguments to use formulas with `cm` cell metadata and `ref` cell …reference

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # openxlsx2 (in development)
 
+## New features
+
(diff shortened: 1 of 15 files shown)


JanMarvin:
* `cm` is pure guesswork. This creates `metadata.xml` and I have no clue what this file is or what it is supposed to do. Either it is a single use case and we have nothing to worry, or it might be highly specific and we create broken xlsx files.
`wb$add_formula(x = 'SUM(ABS(A2:A11))', cm = TRUE)`

* `ref` requires the user to know the size of the reference area in the xml file.
`add_formula(dims = "H1", x = "MMULT(A2:B4, D2:F3)", array = TRUE, ref = "H1:J3")`

JanMarvin:
It took some time, but your [initial draft of a formula vignette](https://github.com/ycphs/openxlsx/issues/299) is now in @zykezero.

PR #604, opened 2023-05-09
add wb_set_sheetview. closes #394

### LINKED ISSUE #394: Get/Set sheetViews
From this [SO](https://stackoverflow.com/questions/74227453/set-excel-worksheet-zoom-level-from-r-openxlsx) provide a function to set sheetViews.

We should also support [more attributes](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.sheetview?view=openxml-2.8.1) while doing this. This includes #142

JanMarvin:
it's probably a bad habit, but the lack of text in the description is originally because I have no idea what all these options do ...


JanMarvin:
This includes changes to `app.xml` which is now created/loaded into `wb$app`.

Issue #105, opened 2022-03-26
removeTable improvements


JanMarvin:
removeTable should work without table. This should be called in deleteData and removeWorksheet.

PR #606, opened 2023-05-10
Update table

JanMarvin:
Parts of this should cover #105
* `wb_remove_tables()` should not throw errors and should only remove the table attributes on a sheet and in the workbook
*  this way it can be added to `delete_data()` or `wb_clean_sheet()` (and maybe the latter should replace the former)
* `wb_update_table()` is a manual way to modify the table xml structure (not sure if this covers every possible xml element and obviously it does not alter the autofilter)
* TODO: `wb_get_named_regions()` as wrapper


R/get-named-regions.R
```diff
@@ -115,9 +116,19 @@ wb_get_named_regions <- function(x) {
...
+    if (is.null(z)) {
+      z <- tb
+    } else {
+      z <- merge(z, tb, all = TRUE, sort = FALSE)
+    }
+
+  }
+
+  if (NROW(z) == 0) {
+    z <- NULL
```
JanMarvin, line 131: remove this

PR #614, opened 2023-05-16
changes the default type to character 

JanMarvin:
Considering #601 there are probably more non default types that can be expressed as numerics than character. Using character might even cause broken XML formats. 

For instance `lubridate` has periods and intervals and I don't want to write handlers for all types possible.


JanMarvin:
have to think about this a bit more

PR #625, opened 2023-05-24
Write Hyperlinks if required. closes #624

### LINKED ISSUE #624: Hyperlink reference not written
Observed on a worksheet with a single external hyperlink:

`wb$worksheets_ref` is `character()` and no `worksheets/_rels/sheet...xml` is written. Can be fixed if the worksheet_ref is manually set to `""`. Most likely some `length()` or `""` comparisson going wrong.

Need to reproduce this locally.

JanMarvin:


JanMarvin:
We write hyperlinks differently. We create a formula and Excel writes the string as is and inserts the hyperlink to the worksheet relationships and links the relationship in the hyperlinks part of the worksheet.

PR #638, opened 2023-06-03
Unsorted cc

JanMarvin:
This PR is removing a few fail safes from our write and update process. Previously we were constructing `cc` as a sorted data frame. Sorted, so that our `cc` order matches the open xml specifications, everythings sorted row wise and not column wise.

This is now skipped. We only sort once, when we write the file and keep our `cc` data frame unsorted. This speeds things up, with the example below I got speed improvements of roughly 20% my Macbook with this 44k x 3 data frame written twice.

There are two downsides
* This impacts a very delicate part of the package
* The improvements are not stable. They vary between 0% and 30%, which makes me wonder what is going on. In addition, this is all based on `milliseconds` 500ms vs 650ms ... It simply might not be worth the risk

```R
beg <- "1900-1-1"
end <- "2022-11-18"

dat <- seq(
  from = as.POSIXct(beg, tz = "UTC"),
  to   = as.POSIXct(end, tz = "UTC"),
  by   = "day"
)

# when writing this creates the 29Feb1900 (#421)
out <- data.frame(
  date = dat,
  chr  = as.character(dat),
  num  = seq_along(dat) - 1
)

### lets go nuts!
library(openxlsx2)

old <- microbenchmark::microbenchmark(
  wb <- wb_workbook()$add_worksheet()$add_data(dims = "D1", x = out)$add_data(x = out),
  times = 25,
  unit = "millisecond"
)

print(old)

### yeeeha!
library(openxlsx2)

new <- microbenchmark::microbenchmark(
  wb <- wb_workbook()$add_worksheet()$add_data(dims = "D1", x = out)$add_data(x = out),
  times = 25,
  unit = "millisecond"
)

print(new)

median(sort(new$time) / sort(old$time) - 1)
```

### DIFF
--- a/tests/testthat/test-save.R
+++ b/tests/testthat/test-save.R
@@ -271,9 +271,10 @@ test_that("write cells without data", {
       f_t = c("", "", "", ""),
       f_ref = c("", "", "", ""),
       f_ca = c("", "", "", ""),
-      f_si = c("", "", "",
-                                                                                                                                  ""), is = c("", "", "", ""), typ = c("3", "3", "3", "3")), row.names = c(NA,
-                                                                                                                                                                                                           4L), class = "data.frame")
+      f_si = c("", "", "", ""),
+      is = c("", "", "", ""),
+      typ = c("3", "3", "3", "3")
+    ), row.names = c(1L, 3L, 2L, 4L), class = "data.frame")
   got <- wb$worksheets[[1]]$sheet_data$cc
   expect_equal(exp, got)
 
--- a/tests/testthat/test-save.R
+++ b/tests/testthat/test-save.R
@@ -258,9 +258,9 @@ test_that("write cells without data", {
 
   exp <- structure(
     list(
-      r = c("B2", "C2", "B3", "C3"),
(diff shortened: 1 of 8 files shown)


tests/testthat/test-save.R
```diff
@@ -271,12 +271,9 @@ test_that("write cells without data", {
...
       f_ref = c("", "", "", ""),
       f_ca = c("", "", "", ""),
-      f_si = c("", "", "", ""),
-      is = c("", "", "", ""),
-      typ = c("3", "3", "3", "3")
-    ),
-    row.names = c(NA, 4L),
-    class = "data.frame")
+      f_si = c("", "", "",
+                                                                                                                                  ""), is = c("", "", "", ""), typ = c("3", "3", "3", "3")), row.names = c(NA,
```
JanMarvin, line 275: hmmm this is symbolic for the level of testing in this PR :smile:

JanMarvin:
Overall I think this might be the way to squeeze out a few moments on large inputs, but I'm not willing to trade 100ms for untested code.

PR #639, opened 2023-06-04
add to_string() to convert variables with labels attribute

JanMarvin:


JanMarvin:
Almost forgot: I added this as another way to approach this issue: https://github.com/co-analysis/a11ytables/issues/93.

This way it should be possible to simply push a vector with labels (e.g. a factor or a labelled vector).

Issue #681, opened 2023-07-11
Thoughts / observations about arguments / fn name consistency.


olivroy:
## Priorities 

* [x] Rename `xlsx_file` argument (may open an issue for that for discussion? 
really could just be `file`, other data import packages like readxl, haven, readr use path or file. I disagree with using `x` as it's more widely used for a vector or for data, not for a file. `file` is used in `wb_load()`  (#708)
> I dont like either, because `xlsx_file` can be both. A file or a workbook (hopefully not a url). That's why I was suggesting some neutral alternative. So far the only consensus is that `xlsx_file` is unacceptable and has to be replaced for the next release (another deprecated argument ...).
> Another possibility would be `wb`.
* [x] Rename `path` -> `file` in `wb_save()`
 

## What I may do

* [x] Rewrite examples - long-term TODO. (#720)
 *  After all, this is a function that is probably never going to be used ever. Creating custom table styles is quite a complex task and I would not recommend it at all. Instead I would try to create a function for custom pivot table styles! :)
* [ ] `wb_row_heights.Rd`, other examples. Separate `create_comment()` from `wb_add/remove()` for consistency.

## Low priorities

* [x] `wbChartSheet` is untested, and I couldn't find much doc.'
  *  underlying driver for `wb_chartsheet()` which can be used with `wb_add_mschart()` and loaded charts. There is not much to test.
* [ ]  Change `params` in `wb_add_conditional_formatting()` to snake_case 

## Done

* [x] ~Snapshot tests should not be git ignored (i.e. `_snaps`), but I guess you had your reasons why~ Oops no snapshot tests, my bad.
* [x] ~?openxlsx2 doesn't really work, I see that you copied your `openxlsx2.R` from `openxlsx.R`~ works after #682 
* [x] `vertAlign` (#693)
* [x] Vignette to document API change between v0.7 and 1.0 (will just need final tweaks after #682)
* [x] ~Export `rowcol_to_dim` to be put in `dims` helper, or use `rowcol_to_dims()` everywhere. and use the internal helper `rowcol_to_dim()` if length is 1?~  (wb_dims() #691 ) Use `wb_dims()` when necessary (#693)
  * ~In some exported functions, the default argument is `rowcol_to_dim(...)`, which is undocumented currently.~ 
* [x]  `ws_*` prefix,-> `wb_` (#693)
* [x] Add an option to warn  each time a camelCase is used to improve the catch (#693)
[...]

PR #691, opened 2023-07-14
provide wb_dims()

JanMarvin:
@olivroy something I quickly typed down in addition to #681, thinking about `ggplot2::aes()` and how clean this approach is. This PR provides `wb_dims()` which can take either `row`/`col`, `rows/cols` or objects that can be coerced via `as.data.frame()`.

``` r
  library(openxlsx2)
  
  # take data frames and col_names & row_names
  wb_dims(mtcars)
  wb_dims(mtcars, col_names = TRUE, row_names = TRUE)
  
  # take vectors
  wb_dims(letters)
  wb_dims(t(letters))
  
  # or time series
  wb_dims(AirPassengers)
  
  # single values
  wb_dims(1)
  
  # named 
  wb_dims(rows =  1:10, cols = 5:7)
  # wb_dims(rows =  1:10, row = 5:7)
  
  # single cell
  wb_dims(row = 5, col = 7)
  
  # two vectors  
  wb_dims(1:10, LETTERS)
  wb_dims(1:10, 1:26)
  
  # in a wb chain
  wb <- wb_workbook()$
    add_worksheet()$
    add_data(x = mtcars)$
    add_fill(
      dims = wb_dims(mtcars, col_names = TRUE),
      color = wb_color("yellow")
    )
```

~~This could be tweaked further, but might provide a simpler helper for users. A name to indicate that it is coming from `openxlsx` could be `wb_dims()`.~~

Please let me know what you think!

[Edit:] Updated


JanMarvin:
`wb_dims()` might be better, because `dims` is just an `s` away from `dim` and a small typo might quickly cause confusion.

olivroy:
Oh yes, I like that, especially with the prefix! 

With this, there wouldn't be a use-case for rowcol_to_dims() in the exported API? 

could inform users of the new syntax. 

But I think it's a step forward for a consistent, streamlined API

JanMarvin:
Yes, this would be a more convenient and possibly simpler to remember function that creates the same output. It's a wrapper around functions.

R/utils.R
```diff
@@ -259,6 +259,70 @@ rowcol_to_dim <- function(row, col) {
...
+  args <- list(...)
+  nams <- names(args)
+
+  col_names <- args$col_names
+  row_names <- args$row_names
+
+  has_cnam <- is.null(col_names)
+  has_rnam <- is.null(row_names)
+
+  if (has_cnam) col_names <- FALSE
```
JanMarvin, line 282: default to true

olivroy:
Hi @JanMarvin ,

after reviewing, `wb_dims()` may be enhanced 

* provide data_only option maybe?
 * For example in a vignette 
```r
wb_dims(mtcars)
"A1:K33"
# I propose:
wb_dims(mtcars, data_only = TRUE)
"A2:K33"
# it could be very useful when applying styling to the data.
```
* Error if only `cols` or `rows` is provided. 

```r
# now
wb_dims(cols = 4)
"A1"
# I suggest 
wb_dims(cols = 4)
"D1"
```

I can file a new issue if it's easier.

JanMarvin:
> wb_dims(x = mtcars, cols = "vs")

like `wb_dims(mtcars, start_cols = "VS")`?

But yes, I though about the same (regarding columns, rownames and data selection). Go for it if you want to improve it. I usually use the cell directly. Open a new issue or a pull request, I currently don't have time for this.

PR #692, opened 2023-07-15
[Draft] read encrypted xlsx files

JanMarvin:
Using a slightly modified version of https://github.com/maharmstone/xlcpp/ to decrypt password protected xlsx files. If a password is provided, we should be able to decrypt the xlsx file. The decrypted file is written to disk and imported.

I do not really like this
* Initially I was aiming to get https://github.com/tfussell/xlnt working. But while they have the [cryptography](https://github.com/tfussell/xlnt/tree/master/source/detail/cryptography) files all sorted in a nice little folder, when I was building, I always ran into a wall. My issue is, that the crypto files are not simply to extract into an independent library.
* ~~xlcpp is aiming for its own workbooks and I built the entire library, just for the decryption part, so this has to be cleaned out.~~
* It depends on ~~libarchive (not sure this is needed) and~~ C++20 features.

```R
  library(openxlsx2)
  fl <- "https://github.com/JanMarvin/openxlsx-data/raw/main/openxlsx2_example_pass.xlsx"
  
  wb <- wb_load(fl, password = "openxlsx2")
```


JanMarvin:
Ah whatever, you have to take my word for it that it builds :)

JanMarvin:
Cleaned the library quite a bit. Builds on Windows, Linux and Mac (on my local one at least), but C++20 is to much

JanMarvin:
Another solution that actually works really well is https://github.com/herumi/msoffice. 

It needs openssl as SystemRequirement and if linked with a static library, encoding and decoding is possible.

PR #693, opened 2023-07-17
Gh issue 681 pt1

JanMarvin:
Work on #681


olivroy:
Impressive work! Thanks !

PR #694, opened 2023-07-17
Stop advertising functions that don't fit in openxlsx2 API

olivroy:
This PR aims to modify the function index https://janmarvin.github.io/openxlsx2/dev/reference/index.html, so that functions that don't fit in the openxlsx2 API are not advertised. 

This will not create breakages, as all functions are still exported.

Adds a bunch of `@keywords internal` to the functions that don't fit in the API. (functions still exported, but not advertised)

~Relocate wb_dims on top of `rowcol_to_dims()` `dims_to_rowcols()`, so it shows up first.~

This PR is minimal. and a overhaul of examples, descriptions links, and deprecation warnings can be done elsewhere (wanted to keep it small)

Edit: this PR got a little larger, because `wb_add_data()` and `write_data()` were documented together. I separated the docs, to make sure that `wb_add_data()` was not marked with `@keywords internal`

Edit2 : I was not able to reproduce interactively, but when running R CMD Check in RStudio's build pane. I get the following error.

```r
   Running the tests in 'tests/testthat.R' failed.
   Last 13 lines of output:
     Error in `basename(sheets$Target)`: object 'sheets' not found
     Backtrace:
         ▆
      1. └─openxlsx2::wb_load(xlsxFile) at test-write.R:135:2
      2.   └─base::basename(sheets$Target)
```


R/utils.R
```diff
@@ -337,6 +264,80 @@ wb_dims <- function(...) {
...
+
+    cols_out <- unique(c(cols_out, cols))
+    rows_out <- unique(c(rows_out, rows))
+  }
+
+  list(cols_out, rows_out)
+}
+#' @rdname dims_helper
+#' @export
+rowcol_to_dims <- function(row, col) {
```
JanMarvin, line 310: I don't like this reordering. Generally code should be read from top to bottom, not in any arbitrary direction.

R/class-workbook-wrappers.R
```diff
@@ -109,9 +109,60 @@ wb_save <- function(wb, path = NULL, overwrite = TRUE) {
...
+#' wb$add_data(sheet = "Formula", x = df)
+#'
+#' #############################################################################
+#' # update cell range and add mtcars
+#' xlsxFile <- system.file("extdata", "openxlsx2_example.xlsx", package = "openxlsx2")
+#' wb2 <- wb_load(xlsxFile)
+#'
+#' # read dataset with inlinestr
+#' wb_to_df(wb2)
+#' write_data(wb2, 1, mtcars, startCol = 4, startRow = 4)
```
JanMarvin, line 164: This example should be rewritten or removed.

JanMarvin (changes requested):
And the R CMD check should be investigated too

R/utils.R
JanMarvin: I know that R is not strict in this regard, but for quite some time something like this would not even compile in c++. `rowcol_to_dims()` is used in `wb_dims()` not the other way around.

olivroy: Ok makes sense, I made this change, to change the ordering in the .Rd file (and in the reference index), but it's not necessary. Reverted

R/class-workbook-wrappers.R
olivroy: Done!

JanMarvin:
Thanks! since we have two interfaces, the lack of exports is something I'm constantly in fear of.

Do you still see the r cmd check, I haven't looked into this yet?

olivroy:
HI @JanMarvin,

regarding the R CMD check, I don't really what is going on. It is failing because of tests only (i.e. only 1 ERROR) It is probably related to a setup on my computer, Windows, R 4.2.3, RStudio 2023.06.0

* This PR basically does nothing.
* It passes when I use Ctrl + Shift + T 
* it passes when I run tests interactively. `devtools::test()`
* It just doesn't pass with RStudio Check command Ctrl + Shift + E
* I see the same thing on main on my machine.


Here is a snippet of the back trace I get.

```r
── Warning ('test-class-comment.R:62:3'): load comments ────────────────────────
internal error in 'unz' code
Backtrace:
    ▆
 1. └─openxlsx2::wb_load(fl) at test-class-comment.R:62:2
 2.   └─utils::unzip(file, exdir = xmlDir)
── Warning ('test-class-workbook-wrappers.R:222:3'): wb_add_drawing is a wrapper ──
internal error in 'unz' code
Backtrace:
    ▆
 1. └─openxlsx2::wb_load(file = fl) at test-class-workbook-wrappers.R:222:2
 2.   └─utils::unzip(file, exdir = xmlDir)
── Warning ('test-class-workbook.R:328:3'): clone worksheet ────────────────────
internal error in 'unz' code
Backtrace:
    ▆
 1. └─openxlsx2::wb_load(fl) at test-class-workbook.R:328:2
 2.   └─utils::unzip(file, exdir = xmlDir)
── Warning ('test-class-workbook.R:737:3'): changing sheet names works with named regions ──
zip file is corrupt
Backtrace:
    ▆
 1. └─openxlsx2::wb_load(filename) at test-class-workbook.R:737:2
 2.   └─utils::unzip(file, exdir = xmlDir)
── Warning ('test-cloneWorksheet.R:9:3'): clone Worksheet with data ────────────
zip file is corrupt
```

I checked on R-devel, nothing seems related.

Only thing that was fixed in R 4.2.2
```r
Files can now be extracted even from very large zip archives ([PR#18390](https://bugs.r-project.org/show_bug.cgi?id=18390), thanks to Martin Jakt).
```

JanMarvin:
Hm, maybe a download didn't succeed? There isn't really a to big file in our tests. You could clone the git repo.

olivroy:
Maybe a couple of `skip_on_cran()` could fix that? but if you never had these problems before, I think it is safe to assume it is related to my machine, and we can ignore until new information convinces us to do otherwise

JanMarvin:
All good on my end

```R
[ FAIL 0 | WARN 0 | SKIP 1 | PASS 1375 ]
```

PR #696, opened 2023-07-18
[Draft] include msoc/cybozu lib for cryptography

JanMarvin:
A lot of files in cybozu, not sure if all are needed. But lets see


JanMarvin:
It works as expected, but atm I'm not comfortable adding a larger library with encryption stuff I'm not really able to maintain to `openxlsx`. The moment CRAN enables a new clang flag, it all might come to a stop. Maybe it should live on in another package `msoc` since it provides OOXML encryption features for more than just `xlsx`.

JanMarvin:
Moved encrypt/decrypt function to external package https://github.com/JanMarvin/msoc

PR #701, opened 2023-07-19
Misc

### LINKED ISSUE #699: Check in add_conditional_formatting is causing many warnings in `rempsyc`
To fix before the next release:

The `rempsyc` package tests are still a hot mess,, but this

```R
rempsyc::cormatrix_excel(mtcars, select = c("mpg", "cyl", "disp", "hp", "carb"), filename = "cormatrix1")
``` 

throws many warnings, due to the newly added check. This should only warn if the input is not constantly increasing by +1. Aka, broken or restarted sequences are not allowed.

https://github.com/JanMarvin/openxlsx2/blob/48bf19f7dce3231b1037f6af171a79b3b519ae6b/R/class-workbook.R#L3918

JanMarvin:
* close #699
* reorganize comment man page (can you double check this @olivroy ?)
* dims for wb_(un)merge_cells
* clean public api for conditional formatting


R/class-comment.R
```diff
@@ -171,15 +171,21 @@ create_comment <- function(text,
   invisible(wbComment$new(text = text, author = author, style = style, visible = visible, width = width[1], height = height[1]))
 }
 
+#' Internal comment functions
+#' 
+#' Users are advised to use [wb_add_comment()] and [wb_remove_comment()]
+#' @name comment_internal
+NULL
 
 #' @name write_comment
```
olivroy, line 180:
```suggestion
```

R/class-comment.R
```diff
@@ -337,9 +343,9 @@ write_comment <- function(
 
 
 #' @name remove_comment
```
olivroy, line 345:
```suggestion
```

olivroy (commented):
Overall, looks good, there is a `#' @name create_comment` at line 87 that should be 
removed in `class-comment.R` too

R/class-comment.R
JanMarvin: Thanks!

PR #702, opened 2023-07-19
Review `wb_dims()`

### LINKED ISSUE #697: Improve `wb_dims()`
Follow-up on #691 

Let me know what you think makes sense.
Imo (unless I missed something)
* wb_dims() should allow only unnamed arguments if there are 2 arguments (numeric) that are rows, cols respectively.
* if using `wb_dims()` on an object, I think it should be named (`x` is probably a good choice.
 * In that case, if `x` is a data.frame.`cols` could allow a character vector for specifying columns by name.
 * Should not support old camelCase, as this is a new function
 * if only `cols` or `rows` is provided, we should error.
 * Possibly correct behaviour for providing an object with `cols` only

Here is my current view / tests. let me know which make sense to you. and I will implement it.
```r
wb_dims(mtcars)
"A1:K33"
# currently 
wb_dims(mtcars, col_names = FALSE)
"A1:K32"
# propose
expect_equal(
 wb_dims(x = mtcars, col_names= FALSE),
"A2:K33"
)
expect_equal(
 wb_dims(x = mtcars, col_names = FALSE, start_col = 2),
"B2:L33"
)
expect_equal(
 wb_dims(x = mtcars, col_names = FALSE, start_row = 2),
"A3:L34"
)
expect_error(wb_dims(mtcars)) # allow unnamed only for rows, cols
# now
wb_dims(cols = 4)
"A1"
wb_dims(rows= 4)
"A1"

# I suggest
expect_error(wb_dims(cols = 4), "Use `start_col = 4` instead. using `cols` alone is ambiguous. Did you mean to specify `x`") 
expect_error(wb_dims(rows= 4), "Use `start_row = 4` instead. using `cols` alone is ambiguous. ")  # not including a hint of using it with an object, as it may not really make sense.

expect_equal(wb_dims(x = mtcars, cols = 4),  wb_dims(x = mtcars, rows = seq_len(nrow(mtcars)), cols = 4))

expect_error(wb_dims(cols = "hp"), "cols must be a numeric, when provided without `x`")
expect_equal(wb_dims(x = mtcars, cols = "hp"),wb_dims(x = mtcars, rows = seq_len(nrow(mtcars)), cols = 4))
expect_error(wb_dims(x = mtcars, cols = c("hp", "vs")), "`wb_dims()` only supports one column at a time.")
# I think it is better to support only 1 column at a time for now, because it wouldn't make sense otherwise. 
# as the use case I see for that would be with formula, or conditional formatting. 

# now
wb_dims(rows = c(1, 3, 4), cols = c(1, 4))
"A1:D4"
# make `wb_dims()` a little more strict as it should only be used to return a single range
expect_error(wb_dims(rows = c(1, 3, 4), cols = c(1, 4)), "wb_dims() should only be used for specifying a single continuous range.")

```

olivroy:
Aims to fix #697 , follow up to #691 

This PR aims to enable the following syntax

EDIT: Updated example to reflect latest API when using `wb_dims()` with `x`.

```r
## add formatting to column names with the help of `wb_dims()`----
wb <- wb_workbook()
wb$add_worksheet("test wb_dims() with an object")
dims_mtcars_and_col_names <- wb_dims(x = mtcars)
wb$add_data(x = mtcars, dims = dims_mtcars_and_col_names)

# Put the font as Arial for the data
dims_mtcars_data <- wb_dims(x= mtcars, select = "data")
wb$add_font(dims = dims_mtcars_data, name = "Arial")

# style col names (with select = "col_names")
dims_column_names <- wb_dims(x = mtcars, select = "col_names")
wb$add_font(dims = dims_column_names, bold = TRUE, size = 13)

# add styling to column "cyl" (the 4th column) (only the data, the default)
# there are many options, but here is the preferred one
# if you know the column index, wb_dims(x = mtcars, cols = 4) also works.
dims_cyl <- wb_dims(x = mtcars, cols = "cyl")
wb$add_fill(dims = dims_cyl, color = wb_color("pink"))

# Add yellow to the vs col, and its column name
dims_vs_and_col_name <- wb_dims(x = mtcars, cols = "vs", select = "x")
wb$add_fill(dims = dims_vs_and_col_name , fill = wb_color("yellow"))

# Cond formatting to the mpg column.
# if specifying cols or rows, the default `select` is "data", 
# if not specifying rows or cols, the default behavior is "x"
wb$add_conditional_formatting(dims = wb_dims(x = mtcars, cols = "mpg"), type = "dataBar")
```
image


R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+    plural_sentence <- ifelse(multi, " is an invalid argument for ", " are invalid arguments for ")
+
+    valid_arg_nams <- paste0("'", choices[choices != ""], "'", collapse = ", ")
+    stop(
+      invalid_arg_nams, plural_sentence, fn_name, ": ", "\n", "Use any of ", valid_arg_nams,
+      call. = FALSE
+    )
+  }
+}
+
```
JanMarvin, line 588: reading order, this should be above

JanMarvin:
Thanks for the pull request. I'll review it later this week. Its already past midnight and it's a bit to complicated for my sleepy head.

R/utils.R
olivroy: Should it be moved to another file, like `asserts.R`?

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+    )
+  }
+  if (n_unnamed_args == 1 && lengt > 1 && !"rows" %in% nams) {
+    message("Assuming the first unnamed argument to be `rows`.")
+    nams[which(nams == "")[1]] <- "rows"
+    names(args) <- nams
+    n_unnamed_args <- length(which(!nzchar(nams)))
+    all_args_unnamed <- n_unnamed_args == lengt
+  }
+  if (n_unnamed_args == 1 && lengt > 1 && "rows" %in% nams) {
```
JanMarvin, line 361: why did you switch the logic here? Why not `!"cols" %in% nams`?

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+#' @param ... construct dims arguments, from rows/cols vectors or objects that
+#'   can be coerced to data frame
 #' @examples
-#' # either vectors
+#' # Provide coordinates
+#' wb_dims()
+#' wb_dims(1, 4)
+#' wb_dims(rows = 1, cols = 4)
+#' wb_dims(start_row = 4)
+#' wb_dims(start_col = 2)
```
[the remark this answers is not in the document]
JanMarvin: Maybe `start_col`/`start_row` was a bad initial idea. I thought of it like `to_col`/`to_row` or `from_col`/`from_row`.

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+#' * `start_row`
+#' * `start_col`
+#'
+#' # Using `wb_dims()` with an object
+#'
+#' * `x` An object (typically a `matrix` or a `data.frame`)
+#' * `start_row` the starting row of `x` (The `dims` returned will be )
+#' * `start_col` the starting column: a single integer, or an Excel column identifier "A", "B" etc.
+#' * `rows` (Which rows? (not fully supported yet
+#' * `cols` a range of column, or one of the column names of `x` (length 1 only accepted)
```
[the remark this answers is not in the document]
JanMarvin: In this case `dims_to_dataframe()` identifies the `;`, but still creates a square matrix and not a list of matrices. And in addition, our styles functions do not handle these. I didn't want to add to much overhead in the functions for conditions the user could simply run in a loop. But of course this would be possible to handle, just ... a new feature.

Maybe: 

```R
foo <- function(dims, bar = wb_color()) {
  if (is.list(dims)) {
     for (i in seq_along(dims)){
        self$foo(dims[[i]], bar = bar)
  }
 ...
}

File: R/utils.R [hunk 5]
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+  }
+  rows_arg <- args$rows
+  #
+  x <- args$x
+  x_has_named_dims <- inherits(x, "data.frame") | inherits(x, "matrix")
+  if (x_has_named_dims && !is.null(rows_arg)) {
+    is_rows_a_colname <- rows_arg %in% colnames(x)
+
+    if (any(is_rows_a_colname)) {
+      stop("`rows` is the incorrect argument. Use `cols` instead.")
```
JanMarvin, line 414: What input would cause to this?

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+    }
+  }
+  if (is.character(rows_arg)) {
+    warning("It's preferable to specify integers indices for `rows`", "See `col2int(rows)` to find the correct index.")
```
JanMarvin, line 418: This does not look right. `col2int()` is supposed to create integer values from spreadsheet column names. `col2int(cols)` or `as.integer(rows)`?

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+      }
+      col_name <- cols_arg
+
+      cols_arg <- which(colnames(x) == cols_arg)
+      message("Transforming col name = '", col_name, "' to `cols = ", cols_arg, "`")
+    }
+  }
 
+  if (!is.null(rows_arg)) {
+    assert_class(rows_arg, class = "integer", envir = parent.frame(n = 2), arg_nm = "rows_arg")
```
JanMarvin, line 448: n == 2 is two up? Like from the current to the parent to the grand parent?

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
-  rows <- rows + srow
-  cols <- cols + scol
+  if (!is.null(cols_arg)) {
+    cols_arg <- col2int(cols_arg)
+    assert_class(cols_arg, class = "integer", envir = parent.frame(n = 2), arg_nm = "cols_arg")
+  }
+
+  srow <- args$start_row %||% 1L
+  srow <- srow - 1L
+  scol <- col2int(args$start_col, allow_null = TRUE) %||% 1L
```
JanMarvin, line 458: I still don't think this `allow_null` is needed here

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+  }
+  # Making sure that at this point, we only cover the case for `x`
+
+  col_names <- col_names %||% x_has_named_dims
+  row_names <- row_names %||% FALSE
+  if (!col_names && row_names && x_has_named_dims) {
+    warning("The combination of `row_names = TRUE` and `col_names = FALSE` is not recommended.",
+      "`col_names` allows to select the region that contains the data only.",
+      "`row_names` = TRUE adds row numbers if the data doesn't have rownames.",
+      call. = FALSE
```
JanMarvin, line 499: Interesting take, but now that I think about, maybe this should be possible. I frequently create data frames with two header columns and write the data straight underneath. The latter, without column names. In such a case it might be ...
But having written this, I don't think I have a use case, where I need to add rownames without another option. Note to self: what does `call.` do?

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+  ncol_to_span <- ncol(x)
+
+  if (col_names && x_has_named_dims) {
+    nrow_to_span <- nrow_to_span + 1
+  }
+  # if without column names and with named dimensions
+  # We will increment the start row by 1.
+  if (!col_names && x_has_named_dims) {
+    srow <- srow + 1
+  }
```
JanMarvin, line 516: Neat!

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+    ncol_to_span <- ncol_to_span + 1
+  }
+
+  # if (row_names) {
+  #   scol <- scol + 1
+  # }
+
+  if (!all(scol >= 0, srow >= 0)) {
+    stop("Internal error. At this point `start_col` and `start_row` should have length 1.")
+  }
```
JanMarvin, line 529: Haha, you really like to catch stuff 👍

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
+    row_span <- srow + seq_len(nrow_to_span)
+    col_span <- scol + seq_len(ncol_to_span)
+  } else if (is.null(rows_arg)) {
+    row_span <- srow + seq_len(nrow_to_span)
+    col_span <- scol + cols_arg + row_names
+  } else if (is.null(cols_arg)) {
+    row_span <- srow + rows_arg + col_names
+    col_span <- scol + seq_len(ncol_to_span)
+  } else {
+    "problem"
```
JanMarvin, line 542: a tad bit vague here 😄

R/utils.R
```diff
@@ -259,84 +259,333 @@ rowcol_to_dim <- function(row, col) {
...
   dims
 }
 
+# It is inspired heavily by `rlang::arg_match(multi = TRUE)` and `base::match.arg()`
+match.arg_wrapper <- function(arg, choices, several.ok = FALSE, fn_name = NULL) {
+  # Check valid argument names
+  # partial matching accepted
+  fn_name <- fn_name %||% "function"
+  # match.arg(arg, choices = choices, several.ok = several.ok)
+  # Using rlang::arg_match() would remove that.
```
JanMarvin, line 565: Just checked, `rlang` apparently only has `utils` as dependency, but unless we really really really ... really cannot avoid this, I don't want any new dependencies.

R/utils.R
JanMarvin: Why? `wb_dims` could be moved to `wb_dims.R` or `wb_to_df()` could be moved to a better named ~~function~~ file and then `wb_function.R` would be a candidate for `wb_dims()`

JanMarvin (commented):
I like the general direction of the pull request and have made a few comments. We might want to think about `start_` arguments and what they are supposed to do.

R/utils.R
olivroy: good catch. I tried putting this in a function at some point, but changed my mind and forgot to revert this change

R/utils.R
olivroy: I often add these conditions for myself  because I want to be bad input later. I want to assume everything is going well after a certain point to avoid potential messy failures.

R/utils.R
olivroy: If you think this function is correct, let's not add it. ;)

R/converters.R
```diff
@@ -20,18 +20,24 @@ int2col <- function(x) {
...
 #' col2int(LETTERS)
-col2int <- function(x) {
-
+col2int <- function(x, allow_null = FALSE) {
+  if (is.null(x)) {
+    if (allow_null) {
+      return(NULL)
+    }
+    warning("`NULL` was provided to `col2int()`, may cause problem, check input carefully.")
+  }
```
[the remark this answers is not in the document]
olivroy: I added this for internal debugging. when I was working on `wb_dims()`, when my conditions were still incorrect. when NULL was passed on to col2int, I ended up with messy results.

I added a allow_null to silence warning in case this was expected, 

similar to `allow_na` in assert_class

R/utils.R
olivroy: Would have to think about this

call. prevents printing 

Warning in wb_dims(.....)

No need for it when you named the function in the meaaage (redundancy)

R/utils.R
olivroy: It is just a personal preference warning 

See one of the tests I corrected

It helped catch this.. 

while it makes sense to have cols as a character, It really doesn't make sene to have rows as a character (even if it's supported)

R/utils.R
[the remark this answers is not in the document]
olivroy: If you replace everywhere start_row, by from_row, you see that it's exactly the mindset I had! Could rename? but that would mean a new param for user (that is not used in the rest of openxlsx2) 
this could be addressed in docs ?

olivroy:
as mentioned in one of the commentd, I see them as `from_row/col` and they are implemented this way

R/utils.R
olivroy: I can add an error. It's not supposed to happen ;)

R/converters.R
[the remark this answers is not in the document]
JanMarvin: Maybe the function should return `integer(0)`. similar to `as.integer()`. I simply don't like the additional argument just for this option.

R/utils.R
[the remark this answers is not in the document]
JanMarvin: Guess it is preferable to have arguments that are self explanatory. So I'd go for it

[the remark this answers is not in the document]
olivroy: Ok, will do, should `start_row`, `start_col` error? or only throw a warning?

R/utils.R
olivroy: I just want to safeguard in case someone messes up the logic.

if someone types,
```r
wb_dims(x = mtcars, rows = "vs")
```

The error message will redirect them to 
```r
wb_dims(x = mtcars, cols= "vs")
```

R/utils.R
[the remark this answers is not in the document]
JanMarvin: Right now it's just in development, therefore we can simply replace. But error is fine too

R/utils.R
JanMarvin: ah okay. Makes sense, unless there's a rowname "vs"

R/utils.R
```diff
@@ -259,85 +259,334 @@ rowcol_to_dim <- function(row, col) {
...
+      invalid_arg_nams, plural_sentence, fn_name, ": ", "\n", "Use any of ", valid_arg_nams,
+      call. = FALSE
+    )
+  }
+}
+#' Helper to specify the `dims` argument.
+#'
+#' @description
+#' `wb_dims()` can be used to help provide the `dims` argument, in the `wb_add_*` functions.
+#' It returns a Excel range (i.e. "A1:B1") or a start like "A2".
```
JanMarvin, line 297: Please change this to something like A1 spreadsheet range or so. I'd prefer to avoid promoting a certain software, even if this is intended. And please add at least a single empty line after a function.

R/utils.R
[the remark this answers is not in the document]
olivroy: Renamed it `from_col` / `from_row` with a special error message if `start_row` / `start_col` are provided.

R/utils.R
olivroy: It currently does not support selecting rows, based on their row name. There could be advantages, but not sure if it's worth the effort...

R/utils.R
olivroy: I removed this warning... But reviewed the API, so that col_names is handled a bit differently

R/utils.R
olivroy: I removed most messages ( I tried to document correctly, so that messages are not required.

R/utils.R
JanMarvin: Fine by me. We do not have to make this overly complicated

R/utils.R
```diff
@@ -259,85 +259,516 @@ rowcol_to_dim <- function(row, col) {
...
+#'
+#' `wb_dims()` with an object has 8 use-cases (they work with any position values of `from_row` / `from_col`),
+#' `from_col/from_row` correspond to the coordinates at the top left of `x` including column and row names.
+#' 1. provide the full grid with `wb_dims(x = mtcars, col_names = TRUE)`
+#' 2. provide the data grid `wb_dims(x = mtcars)`
+#' 3. provide the `dims` of column names `wb_dims(x = mtcars, rows = 0)`
+#' 4. provide the `dims` of row names  `wb_dims(x = mtcars, cols = 0, row_names = TRUE)`
+#' 5. provide the `dims` of a row span `wb_dims(x = mtcars, rows = 1:10)` selects the first 10 rows of `mtcars` (ignoring column namws)
+#' 6. provide the `dims` of data in a column span `wb_dims(x = mtcars, cols = 1:5)` select the data first 5 columns of `mtcars`
+#' 7. provide a column span `wb_dims(x = mtcars, cols = 4:7, col_names = TRUE)` select the data columns 4, 5, 6, 7 of `mtcars` + column names
```
olivroy, line 325: To me, `wb_dims()` with `x` should be this mainly

tests/testthat/test-utils.R
```diff
@@ -46,32 +46,215 @@ test_that("dims to col & row and back", {
...
   )
+  # selecting only content (data)
+  data_content_dims <- "B3:L34"
+  expect_equal(wb_dims_cars(), data_content_dims)
 
-  expect_equal(wb_dims(row = 5, col = 7), "G5")
+  # Selecting a column "cyl"
+  dims_cyl <- "C3:C34"
+  expect_equal(suppressMessages(wb_dims_cars(cols = "cyl")), dims_cyl)
+  expect_equal(suppressMessages(wb_dims_cars(cols = 2)), dims_cyl)
```
olivroy, line 137: Tests to show case it

vignettes/openxlsx2.Rmd
```diff
@@ -185,23 +185,58 @@ In `openxlsx2` functions that interact with worksheet cells are using `dims` as
...
+  add_data(x = mtcars, dims = wb_dims_custom(col_names = TRUE))$
+  add_fill(
+    dims = wb_dims_custom(rows = 1:5),
     color = wb_color("yellow")
+  )$
+  add_fill(
+    dims = wb_dims_custom(rows = 0),
+    color = wb_color("cyan2")
   )
 ```
```
olivroy, line 241: This cool function to make sure that the styling is connected to the data in some way

vignettes/openxlsx2.Rmd
```diff
@@ -185,23 +185,58 @@ In `openxlsx2` functions that interact with worksheet cells are using `dims` as
...
+wb_dims(x = mtcars, col_names = TRUE)
+# only data
+wb_dims(x = mtcars)
+
+# The dims of the values of a column in `x`
+wb_dims(x = mtcars, cols = "cyl")
+# a column in `x` with the column name
+wb_dims(x = mtcars, cols = "cyl", col_names = TRUE)
+# rows in `x`
+wb_dims(x = mtcars)
```
olivroy, line 210: Useful cases

olivroy (commented):
Here are the parts I think are more important. Sorry about the hard to read code in some places..

Maybe another wrapper be even clearer?

I had issues dealing with `add_data(col_names = FALSE)`, so I warn if `x` has column names, but specify `col_names = FALSE`

Edit: thanks a lot for the review!

olivroy:
I will let you fix the lint issues in a separate PR :)

R/utils.R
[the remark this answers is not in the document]
olivroy: Indeed many ranges seems out of scope here. 
But maybe eventually, they could be supported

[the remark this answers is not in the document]
JanMarvin: I just wanted to clarify that this is currently not possible and nothing I have worked on or plan to do in the upcoming weeks.

JanMarvin:
>Edit: thanks a lot for the review!
>
>Edit 2: Feel free to close too. It`s more complicated than I wished it to be. I like the result, and think it can streamline styling and >formatting, but I may be wrong ;)
>
>Edit 3: maybe ask for another review to see if they think this can be useful?

Thanks for the PR and thanks for being open minded! I'll try to review the PR tomorrow or Sunday. I have spent way to much time these last two weeks on `openxlsx2`. Regarding other reviewers, I think this task is up to us :) That's usually the case with software development. Even R core has the issue that nobody reviews/tests draft versions, but everyone complains about "bugged" releases.

Regarding Edit 2 and the increased complexity: after all, I might have stated this already elsewhere, `openxlsx2` provides a base and people can always write custom functions around the base. Therefore a more detailed `wb_dims()` function could be added in a tools or extra package _if_ we decide against this. This way `openxlsx2` does not have to provide everything for everyone.

olivroy:
Thanks! Looking forward to it. 

I forgot to support the use cases, where `x`, `rows` and `cols` are provided. (making sure that `rows` and `cols` are really relative to `x` rather than the workbook. I will possibly warn if  cols = "B:C" for example (and explain what `cols` is and suggest `from_col` instead.

Still buggy with row names. but I will have a look at this later (It's probably just a simple wrong condition somewhere :) )

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+      stop("`rows` is the incorrect argument. Use `cols` instead. Subsetting rows by name is not supported.")
     }
+  }
+  if (is.character(rows_arg)) {
+    warning("It's preferable to specify integers indices for `rows`", "See `col2int(rows)` to find the correct index.")
+  }
 
-    rows <- as.integer(args[[rows]])
-    cols <- col2int(args[[cols]])
+  rows_arg <- col2int(rows_arg)
```
JanMarvin, line 551: Unless `rows_arg` can be a `A1` type column name, this should be `as.integer()`

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+      cols_arg <- which(colnames(x) == cols_arg)
 
-    # wb_dims(data.frame())
-    x <- as.data.frame(x)
-    rows <- seq_len(nrow(x) + col_names)
-    cols <- seq_len(ncol(x) + row_names)
+    }
+  }
+
+  if (!is.null(rows_arg)) {
```
JanMarvin, line 574: can `row_arg` be moved in here? I do not like having code spread in and out of functions. Reading the if condition below this looks like something is missing. If this can only be integer, why do we need the assert afterwards?

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+    assert_class(rows_arg, class = "integer", arg_nm = "rows")
+  }
+
+  if (!is.null(cols_arg)) {
+    cols_arg <- col2int(cols_arg)
+    assert_class(cols_arg, class = "integer", arg_nm = "cols")
+  }
+
+  frow_null <- is.null(args$from_row)
+  srow <- args$from_row %||% 1L
```
JanMarvin, line 584: I used `srow` as short for `start_row`, therefore it should be `frow` now.

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
-      cols <- sel[cols]
-    } else if (valid == 1) {
-      stop("found only one cols/rows argument")
+  if (lengt == 1 && !cond_acceptable_lengt_1) {
+    # Providing a single argument acceptable is only  `x`
+    sentence_unnamed <- ifelse(all_args_unnamed, "unnamed ", " ")
+    stop(
+      "Supplying a single", sentence_unnamed, "argument to `wb_dims()` is not supported.",
+      "\n",
+      "use any of `x`, `from_row` `from_col`. You can also use `rows` and `cols`, You can also use `dims = NULL`"
```
JanMarvin, line 522: `wb_dims(rows = 1)` this throws, even though `rows` is supplied. In this case `cols = 1` could be assumed and vice versa

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+    return(dims)
+  }
 
+  # Fix an error I created
+  # if (frow_null)  {
+  #   srow <- srow + 1L
+  # }
+  # if (fcol_null) {
+  #   scol <- scol + 1L
+  # }
```
JanMarvin, line 631: commented code should be removed same with commented messages. either include them in some `if (debug) {}` statement or remove them entirely (my preference).

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+  if (x_present && !col_names && x_has_named_dims && !cnam_null) {
+    if (x_has_colnames) {
+      warning("`x` has column names. Yet, you are asking for `col_names = FALSE`.",
+        "\n ",
+        "\n Consider supplying `x = unname(`input`)`, or use `wb_dims()` without `x` to ensure no errors with `col_names = FALSE`",
+        call. = FALSE
+      )
+    } else {
+      # message("`x` doesn't have col names. assuming there is no name. Supply `col_names = TRUE` only to select rows + column name.")
+    }
```
JanMarvin, line 669: remove `else`?

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+  }
 
-  if (length(rows) == 1 && length(cols) == 1) {
-    # A1
-    dims <- rowcol_to_dim(rows, cols)
+  if (!cnam_null && !x_has_named_dims) {
+    stop("Supplying `col_names` when `x` is a vector is not supported.")
```
JanMarvin, line 673: Unfortunately having something like this might be required. I do not recall from the top of my head, but there is a case where we write characters with column names. But it might be with transposed vectors, which are matrices and not vectors anymore. Anyhow it should not be an error, maybe a message that col_names with a vector was selected.

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+
+  if ((!is.null(rows_arg) && !col_names) && !is.null(cols_arg) && row_names && x_has_named_dims) {
+    warning("The combination of `row_names = TRUE` and `col_names = FALSE` is not recommended.",
+      "unless supplying `cols` and/or `rows`",
+      "`col_names` allows to select the region that contains the data only.",
+      "`row_names` = TRUE adds row numbers if the data doesn't have rownames.",
+      call. = FALSE
+    )
+  }
+  if (!frow_null && identical(srow, -1L)) {
```
JanMarvin, line 687: we could enforce `max(1, srow)`/`max(1, scol)`

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+  }
+
+  if (row_names && !identical(cols_arg, 0L)) {
+    # Will not interact with row_name, unless `cols = 0`
+    scol <- scol + 1L
+  }
+
+  if (identical(scol, 0L) || identical(srow, 0L)) {
+    is_ok_if_from_col_is_zero <- fcol_null | isFALSE(row_names) | x_has_named_dims
+    is_ok_if_from_row_is_zero <- frow_null | isFALSE(col_names) | x_has_named_dims
```
JanMarvin, line 761: `||` just to be on the safe side

JanMarvin (commented):
Gave it a larger review. I want to see case 2 as the default case. So this should be returned

```R
wb_dims(x = mtcars) # like col_names = TRUE
#> "A1:K33"
```

To match our `wb_add_data()` way of handling data frames. Otherwise clean it a bit and tick ready for review and we can merge.

The man page has a few large code blocks without any formatting. I did not really try to understand why we need `match.arg_wrapper()`, but I don't have to understand everything :smile:

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+#' # dims of all the data of mtcars.
+#' wb_dims(x = mtcars, col_names = FALSE)
+#'
+#' # dims of the column names of an object
+#' wb_dims(x = mtcars, rows = 0, col_names = TRUE)
+#'
+#' ## add formatting to column names with the help of `wb_dims()` ====
+#' wb <- wb_workbook()
+#' wb$add_worksheet("test")
+#' wb$add_data(x = mtcars, dims = wb_dims(x = mtcars))
```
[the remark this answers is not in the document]
olivroy: I agree that this may not be the right thing
Imo `wb_dims()` with `x` works better and makes more sense with column names

# work with mtcars full
wb_dims(x = mtcars, col_names = TRUE)
"A1:K33"
# work with mtcars data
wb_dims(x = mtcars)
 "A2:K33"
# work with mtcars without column nameds
wb_dims(x = unname(mtcars), col_names = FALSE)



But do you agree with this ?
```r
wb_dims(x = mtcars, cols = "cyl")
#>  "B2:B33"
```

Because when applying conditional styling, you usually don't want to touch the column name.

And use this code this
```r
wb_dims(x = mtcars, cols = "cyl", col_names = TRUE)
#>  "B1:B33"
```

R/utils.R
olivroy: Well, I am warning against this, but still allowed technically, maybe I should just error instead of warn, so as.integer could be used.

FWIW, the warning helped me catch warnings in these tests. See 8be9dd8697b7fd13a24be572d6313c02117a6b18, otherwise would have errored I think.

R/utils.R
olivroy: It gave the wrong result in the previous version...
I think that the user should really use
```r
wb_dims(from_row = 1)
```

```r
wb_dims()
wb_dims(NULL)
wb_dims(1, 1)
wb_dims(1, 2)
wb_dims(cols =1, 2)
```
are all acceptable alternatives

R/utils.R
olivroy: Related to 8be9dd8697b7fd13a24be572d6313c02117a6b18. In some tests or examples, characters were used to specify rows.. At some point later, it could be changed, but until rows/start rows can be characters, we still need `col2int()` I think..

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+
+  if (!x_present) {
+    if (!cnam_null || !rnam_null) {
+      stop("`row_names`, and `col_names` should only be used if `x` is present.")
+    }
+  }
+  rows_arg <- args$rows
+  #
+  x <- args$x
+  x_has_named_dims <- inherits(x, "data.frame") | inherits(x, "matrix")
```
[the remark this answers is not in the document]
olivroy: is.null(dimnames(x)) maybe? but this would be NULL if a matrix doesn't have row or colnames

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
   # we will always return something like "A1"
   stringi::stri_join(min_col, min_row)
 }
-#' @rdname dims_helper
-#' @param ... construct dims arguments, from rows/cols vectors or objects that can be coerced to data frame
+
+# It is inspired heavily by `rlang::arg_match(multi = TRUE)` and `base::match.arg()`
```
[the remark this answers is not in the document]
olivroy: Only inspired because it's just that it doesn't allow partial matching compared to base R.
rlang is licensed with MIT too. I can remove heavily though

R/utils.R
```diff
@@ -259,85 +259,560 @@ rowcol_to_dim <- function(row, col) {
...
+#' \dontrun{
+#' wb <- wb_workbook()
+#' # Supplying dims using x
+#' dims_column_names <- wb_dims(x = mtcars, rows = 0)
+#' wb$add_font(dims = dims_column_names, bold = TRUE)
+#'
+#' # Finally, to add styling to column "cyl" (the 4th column)
+#' # there are many options, but here is the preferred one
+#' # if you know the column index, wb_dims(x = mtcars, cols = 4) also works.
+#' dims_cyl <- wb_dims(x = mtcars, cols = "cyl")
```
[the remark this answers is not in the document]
olivroy: Would the following be more informative
          "Supplying multiple column names is not supported by the `wb_dims()` helper, ",
          "use the `cols` with a range instead of `x` column names.",
          "\n Use a single `cols` at a time with `wb_dims()`"

R/utils.R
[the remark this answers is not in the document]
olivroy: also, I am using this to validate arguments and produce an informative error message for users. And to do this, I use 30 lines, I felt like it made the already heavy `wb_dims()` function heavier. Also, it can be used in other places, where a user should supply values or arguments that need to be in a certain set

R/utils.R
olivroy: I would be careful before loosening up, (I want to have tests for use cases as much as possible !) But I wouldn't mind, when a use case arises

R/utils.R
olivroy: That would be a good idea. I just want to enforce `rows` `cols` = 0 as a special case with `x`, but it's been a little tricky :)

olivroy:
Thanks again for the review. I didn't touch the logic in my last few commits. But I wanted to showcase my last example
```r
## add formatting to column names with the help of `wb_dims()`----
wb <- wb_workbook()
wb$add_worksheet("test wb_dims() with an object")
dims_mtcars_and_col_names <- wb_dims(x = mtcars, col_names = TRUE)
wb$add_data(x = mtcars, dims = dims_mtcars_and_col_names)
#'
# Put the font as Arial for the data
dims_mtcars_data <- wb_dims(x = mtcars)
wb$add_font(dims = dims_mtcars_data, name = "Arial")
#'
# Style col names as bold using the special `rows = 0` with `x` provided.
dims_column_names <- wb_dims(x = mtcars, rows = 0)
wb$add_font(dims = dims_column_names, bold = TRUE, size = 13)
#'
# Finally, to add styling to column "cyl" (the 4th column) (only the data)
# there are many options, but here is the preferred one
# if you know the column index, wb_dims(x = mtcars, cols = 4) also works.
dims_cyl <- wb_dims(x = mtcars, cols = "cyl")
wb$add_fill(dims = dims_cyl, color = wb_color("pink"))

# Mark a full column as important(with the column name too)
wb_dims_vs <- wb_dims(x = mtcars, cols = "vs", col_names = TRUE)
wb$add_fill(dims = wb_dims_vs, fill = wb_color("yellow"))
```
image

JanMarvin:
Nice example. Still, I don't like changing col_names from our default value for wb_dims(). You may drop it if x and cols are selected or can add some additional selection function. Something like "top, left, right, bottom, data" if you want to pick something explicitly. But picking the data part without the columns per default is somewhat counter intuitive.

R/utils.R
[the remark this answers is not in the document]
JanMarvin: Yes, I agree with the latter part. If a column is selected we can default to ignore the column name row unless it is explicitly requested. Though the default should be to include the column name with data frames.
I just saw that you proposed defaulting to `col_names = FALSE` in your initial proposal, sorry for not seeing it in the first place and now being picky.

R/utils.R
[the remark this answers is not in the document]
JanMarvin: I have to think about it, anyhow not for this PR :)

R/utils.R
JanMarvin: This old test might be two years old and from a time where our internals were in a process of being entirely rewritten by me.
But, start row should under no circumstances require `col2int()` if some character ends up here it should not be a column name.

R/utils.R
JanMarvin: OK, fine by me, as stated, we don't have to make this more complicated than it already is.

R/utils.R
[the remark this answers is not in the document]
JanMarvin: I just wanted to make sure that we are not directly copying a function. Otherwise we should have added some license header for `rlang`. If it's just inspired that's fine by me.

People use to say "the pathway to hell is plastered with good intentions". Now we have this function here, that isn't used anywhere else but in `wb_dims()` and `wb_dims()` is already bloated. And we need changes to col2int() and assert_class() only to create a function constructing a simple A1 cell reference. And please be aware, I'm not saying these changes aren't justified, but I want to make sure that they are not just introduced because of either laziness (now or by previous developers) or lack of knowledge of the `openxlsx2` internals. Because that's what happened to `openxlsx`, where e.g. multiple functions for `col2int()` were introduced over the years (even in R and C++), because everyone always assumed that they were needed and nobody considered that maybe the function they were writing was already around.

olivroy:
I have thought of a new and simpler implementation. I should push tomorrow.

Instead of relying on rows = 0 , I added an argument

select = "col_names"
I was able to simplify it a lot. may open a new PR as this one is getting cluttered

olivroy:
HI @JanMarvin, this is ready for review!
I have simplified a lot the logic, added 2 helpers to concentrate most checks in a single place.
changed the default back 
Updated the vignettes to showcase the latest API.
Added a select argument:

* "x" select data + col names and row names (default if rows or cols is not provided
* "data" selects data only (default if cols or rows is provided)
* "col_names" select col names only (I am not yet supporting select a single column name)
* "row_names" selects row names only  (I am not yet supporting selecting a single row name)


this approach gets rid of the `rows = 0` syntax! (I wrote it, but I found it confusing, so I can't imagine someone who didn't write it)

Ideally, it would be great to start a new PR with a clean state, but I don't know how to do that...)

R/utils.R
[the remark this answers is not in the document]
olivroy: Hi. 

The changes to col2int are mostly related to base R's inconsistencies...

as.integer(NULL) returns integer(0), but nrow(NULL) returns NULL.

NULL is much easier to handle than integer(0) hence the small change.

regarding the now 3 functions only used by wb_dims(), I wouldn't mind putting them back inside wb_dims() if you'd like. 
sometimes, it's just that more scoped functions are easier.

I really don't like base R error messages, and the fact that it's failing silently sometimes.
```r
match.arg(c("rows", "whatever", "cols", "col"), c("cols, "rows"), several.ok = T)
"rows" "cols" "cols" 
```
hence the need for the wrapper..

assert_class changes are also because when I struggle at first, an internal error was thrown by assert_class(), because `sx` was `character(0)`, which was reallly hard to debug...

R/utils.R
```diff
@@ -259,82 +259,500 @@ rowcol_to_dim <- function(row, col) {
...
+#' if they are `rows`, `cols`, for example `wb_dims(1:4, 1:2)`, that will return "A1:B4".
+#'
+#' `wb_dims()` can also be used with an object (a `data.frame` or a `matrix` for example.)
+#' All parameters are numeric unless stated otherwise.
+#'
+#' # Using `wb_dims()` without an `x` object
+#'
+#' * `rows` / `cols` (if you want to specify a single one, use `from_row` / `from_col`)
+#' * `from_row` / `from_col` the starting position of the `dims`
+#'    (similar to `start_row` / `start_col`, but with a clearer name.)
```
JanMarvin, line 393: Also the intention is different at least in the way I see it. With `start_` you are reading from a column downwards (all 1M rows or 10k rows). With `from_` you are placing a cell range to a specific part of the worksheet.

R/utils.R
```diff
@@ -259,82 +259,500 @@ rowcol_to_dim <- function(row, col) {
...
+#'
+#' ```R
+#' wb_dims_cars <- function(...) {
+#'   wb_dims(x = mtcars, from_row = 2, from_col = "B", ...)
+#' }
+#' # using this function
+#' wb_dims_cars() # full grid (data + column names)
+#' wb_dims_cars(select = "data") # data only
+#' wb_dims_cars(select = "col_names") # select column names
+#' wb_dims_cars(cols = "vs") # select the `vs` column
```
JanMarvin, line 426: maybe consider aligning on the comment so that readability is a bit increased

R/utils.R
```diff
@@ -259,82 +259,500 @@ rowcol_to_dim <- function(row, col) {
...
+#' @param select A string, one of the followings.
+#'    it improves the selection of various parts of `x`
+#'    One of "x", "data", "col_names", or "row_names".
+#'    `"data"` will only select the data part, excluding row names and column names  (default if `cols` or `rows` are specified)
+#'    `"x"` Includes the data, column and row names if they are present. (default if none of `rows` and `cols` are provided)
+#'    `"col_names"` will only return column names
+#'    `"row_names"` Will only return row names.
+#'
+#' @return A `dims` string
+#' @export
```
JanMarvin, line 468: The examples are a bit out of hand. In a follow up we can move them to a vignette (openxlsx2.Rmd) or something else.

R/utils.R
```diff
@@ -259,82 +259,500 @@ rowcol_to_dim <- function(row, col) {
...
+#'
+#' # Mark a full column as important(with the column name too)
+#' wb_dims_vs <- wb_dims(x = mtcars, cols = "vs", select = "x")
+#' wb$add_fill(dims = wb_dims_vs, fill = wb_color("yellow"))
+#' wb$add_conditional_formatting(dims = wb_dims(x = mtcars, cols = "mpg"), type = "dataBar")
+#' # wb_open(wb)
+wb_dims <- function(..., select = NULL) {
   args <- list(...)
-  nams <- names(args)
+  lengt <- length(args)
```
JanMarvin, line 525: I'd prefer `len`, but ignore that 😄

R/utils.R
```diff
@@ -259,82 +259,500 @@ rowcol_to_dim <- function(row, col) {
...
+  n_unnamed_args <- length(which(!nzchar(nams)))
+  all_args_unnamed <- n_unnamed_args == lengt
+  # argument dispatch / creation here.
+  # All names provided, happy :)
+  # Checking if valid names were provided.
 
-  if (srow_null) srow <- 0 else srow <- srow - 1L
-  if (scol_null) scol <- 0 else scol <- col2int(scol) - 1L
+  if (n_unnamed_args > 2) {
+    stop("only `rows` and `cols` can be provided without names. You must name all other arguments.")
```
JanMarvin, line 553: lower case `only`

R/utils.R
```diff
@@ -259,82 +259,500 @@ rowcol_to_dim <- function(row, col) {
...
+    is.atomic(args[[1]]) || any(nams %in% c("rows", "cols"))
 
-  x <- args[[1]]
-  exp_name <- inherits(x, "data.frame") || inherits(x, "matrix")
+  if (nams[1] == "" && !ok_if_arg1_unnamed) {
+    stop(
+      "The first argument must either be named or be a vector.",
+      "Providing a single named argument must either be `from_row`, `from_col` or `x`."
+    )
+  }
```
JanMarvin, line 569: Please add empty lines after a closing `}`. This already is a huge function, these blank lines will not hurt us

tests/testthat/test-utils.R
```diff
@@ -45,33 +45,217 @@ test_that("dims to col & row and back", {
   expect_equal(exp, got)
 
 })
+test_that("`wb_dims()` works/errors as expected with unnamed arguments", {
 
-test_that("wb_dims() works", {
+  # Acceptable inputs
+  expect_equal(wb_dims(), "A1")
+  expect_equal(wb_dims(NULL), "A1") # to help programming with `wb_dims()` maybe?
```
JanMarvin, line 52: This is a bit strange, but ok

tests/testthat/test-utils.R
```diff
@@ -45,33 +45,217 @@ test_that("dims to col & row and back", {
...
 
-  expect_equal(wb_dims(letters), "A1:A26")
+  # This used to return A1 as well.
+  expect_error(wb_dims(2), "Supplying a single unnamed argument is not handled")
+  expect_error(wb_dims(mtcars), "Supplying a single unnamed argument")
 
-  expect_equal(wb_dims(t(letters)), "A1:Z2")
 
-  expect_equal(wb_dims(1), "A1")
+  skip("lower priority, but giving non-consecutive rows, or cols should error in `wb_dims()`")
```
JanMarvin, line 72: Why should this be skipped?

JanMarvin (commented):
Thanks, my only regret is that this `wb_dims()` function is growing bigger and bigger. I didn't want to compete with `cellranger`.

Minor comments, otherwise if should be fine.

R/class-workbook-wrappers.R
```diff
@@ -1903,8 +1903,7 @@ wb_remove_filter <- function(wb, sheet = current_sheet()) {
 #'   "t" = as.POSIXct("2016-01-01") + -5:5 * 10000
 #' )
 #' wb$add_data_table(2, x = df)
-#' wb$add_data_validation(2,
-#'   col = 1, rows = 2:12, type = "date",
+#' wb$add_data_validation(2, dims = "B2:B12", type = "date",
```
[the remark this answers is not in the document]
olivroy: Oops, sorry, will change it back

R/utils.R
olivroy: I changed it ;), just replace all lengt -> len, then  replace all lenh -> length ;) 2 seconds

tests/testthat/test-utils.R
olivroy: It still doesn't error / warn if users provide `wb_dims(c(1, 3, 4), 1:4)`

tests/testthat/test-utils.R
olivroy: Would you prefer something else? I just wanted to have edge cases to work anyway..

R/utils.R
olivroy: I tried some reorg, but could always be done in a follow-up.

R/utils.R
olivroy: Ok, do you have places where this is implemented elsewhere?

I used your initial ideas from `wb_dims()`, but maybe we can continue the discussion on the issue?

R/utils.R
```diff
@@ -259,82 +259,500 @@ rowcol_to_dim <- function(row, col) {
...
+
+  x_has_colnames <- !is.null(colnames(args$x))
+
+  if (x_has_colnames && !is.null(args$rows) && is.character(args$rows)) {
+    # Not checking whether it's a row name, not supported.
+    is_rows_a_colname <- args$row %in% colnames(args$x)
+
+    if (any(is_rows_a_colname)) {
+      stop(
+        "`rows` is the incorrect argument in this case\n",
```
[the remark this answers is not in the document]
olivroy: I like them, especially when they are longer. I find them easier to read.
I know you are not a `library(tidyverse)` guy. but I think that improving error messages has been a high priority lately, and has made debugging easier for me. https://style.tidyverse.org/error-messages.html#error-details

One of the central part is that they split long error messages on multiple lines.

tests/testthat/test-utils.R
JanMarvin: I'd assume that it returns NULL or even stop, because otherwise the wb function will complain. IMHO this could let the user know that something is wrong on their side

R/utils.R
JanMarvin: Sure. Not sure what you mean with elsewhere. The initial proposal was a quick draft to have something nicer than `rowcol_to_dims()`. I most likely haven't thought about all corner stones. I usually like to get things rolling and fine-tuning can happen afterwards.

tests/testthat/test-utils.R
olivroy: done!

R/utils.R
olivroy: I mean in other functions of openxlsx2.

But I'll be happy to do modifications afterwards too!

olivroy:
I put it as error, can always change behavior should need arise! but imo, it's better to be explicit! 

I think that wb_dims() could make code easier to read, but using it withoutarguments is just mysterious. no need for that ;) it is easy to type wb_dims(1, 1)

R/utils.R
```diff
@@ -259,79 +259,507 @@ rowcol_to_dim <- function(row, col) {
...
+#'    `"x"` Includes the data, column and row names if they are present. (default if none of `rows` and `cols` are provided)
+#'    `"col_names"` will only return column names
+#'    `"row_names"` Will only return row names.
+#'
+#' @return A `dims` string
+#' @export
 #' @examples
-#' # either vectors
+#' # Provide coordinates
+#' wb_dims()
```
olivroy, line 472:
```suggestion
```

man/wb_dims.Rd
```diff
@@ -0,0 +1,159 @@
...
+wb_dims_cars(select = "col_names") # select column names
+wb_dims_cars(cols = "vs") # select the `vs` column
+}\if{html}{\out{</div>}}
+
+It can be very useful to apply many rounds of styling sequentially.
+}
+
+\examples{
+# Provide coordinates
+wb_dims()
```
olivroy, line 107:
```suggestion
```

olivroy (commented):
remove outdated example

JanMarvin:
> I think that wb_dims() could make code easier to read, but using it without arguments is just mysterious. no need for that ;) it is easy to type wb_dims(1, 1)
 I didn't think of that, but yes `wb_dims()` is a helper just like `wb_color()`. If `A1` is requested, there are a few possibilities.

R/converters.R
```diff
@@ -16,22 +16,26 @@ int2col <- function(x) {
...
 #' @param x A character vector
+#' @return An integer column label (or `NULL` if `x` is `NULL`)
 #' @export
 #' @examples
 #' col2int(LETTERS)
 col2int <- function(x) {
-
+  if (is.null(x)) {
+      return(NULL)
+    }
```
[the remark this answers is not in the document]
JanMarvin: It's still there, I'll fix it.

R/utils.R
JanMarvin: Ah, got it. We have `start_` in `wb_to_df()` and it should (soon) be deprecated in all other functions in favor of `dims`. Previously we had competing arguments for selections like `dims`, `cols/rows`, `col/row` and `startCol/startRow`. Now we have only one and soon a nice wrapper for `dims` that can provide the other variants.

We still have remaining `start_` arguments in `wb_add_data()` and its derivatives. These can be removed, but we can do that after the merge. This way we'll lose another bunch of confusing arguments. But there we have to be careful, might require some under the hood conversions, because it will be a last minute change under the hood. We can discuss this somewhere else.

R/utils.R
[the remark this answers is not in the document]
JanMarvin: Thanks, I guess we could see if we want to extend it in the future. After all it would simply be a tiny loop over select arguments and `paste(dims, collapse = ";")` at the end. But sometimes you put a lot of effort into something and in the end it will only be used by you and only while you are developing the feature. We can wait until feature requests appear (in all honesty I don't want to review `wb_dims()` related stuff in the very near future :wink: ).

R/utils.R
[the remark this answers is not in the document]
JanMarvin: Another || case

JanMarvin:
I'll squash merge once the CI is green. Thank you for sticking to it and for the PR!

Hope you had at least a tiny bit of fun and maybe learned a bit or two about `openxlsx2` :sunglasses:

olivroy:
Tjanks for letting me!
More work than I imagined.  thanks for your thorough revire. 

Hopefully, it can help users! 

It will help me ar least

PR #708, opened 2023-07-26
Gh issue 681 pt2

JanMarvin:
In addition to #681. 

* remove deprecated function `read_sheet_names()` (it looks horrible and was deprecated in 0.7?)
* make `wb_get_sheet_name()` internal. There is a wrapper function `wb_get_sheet_names()` which is just an `s` away.
* change `xlsx_file` to `file`, another option I was considering is `wb`. Because file can be a file or a workbook and `wb` kinda reflects that. Thoughts?


sjewo:
My code broke because of the removed `read_sheet_names()` function and it took some time to find out what the problem was.

So I leave this here for further reference:
```
# replacement for read_sheet_names
wb_get_sheet_names(wb_load(filepath))
```

JanMarvin:
Thanks!

PR #717, opened 2023-08-05
Documentation improvements

olivroy:
This is a couple of documentation improvements. Addresses #681

* Merges some add / remove documentation together in a single .Rd file
* get/set together
* Removes duplicated entries in the pkgdown index
* Removes redundant `@seealso` since there is already crossreferences with the `@family` tag
* Rename `R/openxlsx2.R` -> `R/openxlsx2-package.R`
* Remove some mentions to `write_data()`, `write_datatable()` in favour of `wb_add_data()`, `wb_add_data_table()`
* `workbook_grouping.Rd` `wb_grouping.Rd`
* Rename helper functions so that test file and R file align
* Move get_cell_refs tests in test-converters.R as get_cell_refs lives in `converters.R`

This may not be as focused or thorough as I wished it to be, but these are leftovers that I think can be useful. 

Edit: maybe `wb_set_col_widths()` could be `wb_add_col_widths()`?


R/class-workbook-wrappers.R
```diff
@@ -1796,20 +1787,20 @@ wb_remove_named_region <- function(wb, sheet = current_sheet(), name = NULL) {
...
+#' Add or remove excel column filters to a worksheet
 #'
+#' Adds filters to worksheet columns, same as filter parameters in write_data.
+#' [wb_add_data_table()] automatically adds filters to first row of a table.
+#' NOTE Can only have a single filter per worksheet unless using tables.
 #' @param wb A workbook object
-#' @param sheet A name or index of a worksheet
+#' @param sheet
+#'   * In `wb_add_filter()` A single name or index of a worksheet.
+#'   * In `wb_remove_filter()`, a vector of names/indices of worksheet
```
JanMarvin, line 1800: Thanks for the PR! Looking good, minor nitpicking: This should be the same as everywhere else. The mistake was in the never updated documentation.

JanMarvin:
> Edit: maybe wb_set_col_widths() could be wb_add_col_widths()?

This might be a good idea indeed.

R/class-workbook-wrappers.R
olivroy: I have changed it, since I realized the .Rd file didn't really render as expected. 

But do you want me to update the docs for other `wb_remove_*` functions? As of now, I haven't touched the documentation for other `wb_remove_*` functions

JanMarvin: Guess wherever we use a sheet waiver the same text should work. It just looks like the arguments take different input.

olivroy: Ok, still unsure of how to implement it.

JanMarvin:
Thanks for the pull requests! Your really helping in a long time overlooked area of the package

PR #718, opened 2023-08-06
Improve `wbWorkbook` documentation.

olivroy:
Addresses #681.
The reason why the documentation did not render properly for `wbWorkbook` is because of issues in roxygen2. namely r-lib/roxygen2#1148 and r-lib/roxygen2#1504

The documentation for `wbWorkbook` and `wbChartSheet looks much better after this PR. 

I looked at how processx documents its `process` R6 Class.
https://github.com/r-lib/processx/blob/main/man/process.Rd and https://github.com/r-lib/processx/blob/main/R/process.R for the record.

Edit: the changes in the R files could be reverted once the roxygen2 fixes are made.

The `wbWorkbook` docs could be even cleaner if some documentation of fields was removed ? i.e. see methods first? Also could add links to vignettes on how to use it, although, it is mentioned in many places, how to use the R6 methods, `wb$add` etc. (for amother PR or issue if relevant)

### DIFF
--- a/R/class-chart-sheet.R
+++ b/R/class-chart-sheet.R
@@ -1,6 +1,7 @@
 
 #' R6 class for a Workbook Chart Sheet
 #'
+#' @description
 #' A chart sheet
 #'
 #' @export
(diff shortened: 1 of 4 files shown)


JanMarvin:
Thank you!

PR #720, opened 2023-08-06
Review examples

olivroy:
This is a minimal review of examples, lmk if there are other functions that would be priorities for 0.8 release. I don't mind doing more. 
Just wanted to present a little something as some examples were using internal / old arguments or functions, minor reorgs as well, don't mind reverting some, as some may reflect my personal preferences.

Address #691


JanMarvin:
looking good. I might squeeze in another round of cleanups for a certain spreadsheet software name, but should be merged soon.

PR #722, opened 2023-08-06
Rename `set_col_widths()` -> `add_col_widths()`, `set_row_heights()` -> `add_row_heights()`

olivroy:
Here is what I typed quickly. just felt it made the function names more consistent as most `add` functions were related to data in a worksheet, while `set` functions seem to be more related to the full workbook.
Feel free to close if this is not something you think is worthwhile1

```r
  wb <- wb_workbook()$
    add_worksheet()$
    add_col_widths(cols = 1:10, width = (8:17) + .5)$
    add_data(x = rbind(8:17), col_names = FALSE)
```

Edit: I did the reordering to keep row-related functions together and cols-related functions together, hence the larger diff.

Edit 2: for reading order, I don't mind moving the `set` functions to `openxlsx2-deprecated.R`


JanMarvin:
I don't really want this. I guess it breaks the getter/setter logic and I think that we are so far mostly done with breaking changes for 1.0. Something like this, a greater review of functions and function pairs, could happen with respect to 2.0 and probably only once consensus has been reached as to what is supposed to happen.

Otherwise the next person comes along and wants to rename it to wb_modify_row_height()/wb_modify_column_heitght(). And we end up in a never ending spiral of renaming the same old functions. After all, let's face it, people write their code once and use what they find on the internet. Otherwise `openxlsx` wouldn't be used and recommended anymore.

olivroy:
I agree this would require more thought too!

PR #748, opened 2023-08-21
fix wb_set_header_footer(). closes #747

### LINKED ISSUE #747: Special tags of `set_header_footer()` give me a corrupt file
Running the examples of `wb_set_header_footer()` give me a corrupt file when using the special tags.

I am running Microsoft Excel 365 on Windows 10, with R 4.2.3 in a French locale..

```r
wb <- wb_workbook()

# Add example data
wb$add_worksheet("S1")$add_data(x = 1:400)
wb$set_header_footer(
sheet = 1,
header = c("&[Date]", "ALL HEAD CENTER 2", "&[Page] / &[Pages]"),
footer = c("&[Path]&[File]", NA, "&[Tab]"),
first_header = c(NA, "Center Header of First Page", NA),
first_footer = c(NA, "Center Footer of First Page", NA)
)
wb$open()
```

If you can't reproduce, I will provide more details!

It's written Illegal character name

Edit: I encounter the same `wb$add_worksheet()`  and `wb$set_header_footer()`

JanMarvin:


olivroy:
This seems to be working with `wb_set_header_footer()` but still bugging in `wb_add_workseet()`.

Maybe it would be an option to leave it out of `wb_add_worksheet()` in favour of `wb_set_header_footer()`?

JanMarvin:
Could you check #751 (I copied code around, wasn't even aware that it is possible to set header/footer from `add_worksheet()` relict from `openxlsx`. A better fix would be to provide some `worksheet$set_header_footer()` function, but to lazy.

JanMarvin:
Darn, it is not even passing our own tests ... 😞

olivroy:
But it's not a feature that is very useful  (even less for `wb_add_worksheet()`) It is just the problem of providing working examples

JanMarvin:
I agree, another one to ditch in the future. There are many things that were not obvious to me when I started this little adventure. For instance the strange `create_comment()` and `write_comment()` logic. It might have made sense earlier, but right now I would simply use `wb_add_comment(dims = "A1", comment = "foo")`, but 2 years ago I wasn't aware (most likely everything was in a broken state :)).

Now I live with it and add it to my imaginary TODO list for the next major release after `1.0`. After all fixing things like this takes time as well, testing, developing etc. and rest assured, I haven't written a xlsx comment with `openxlsx` or `openxlsx2` in my life apart from development and probably won't do so.

olivroy:
Ooh, and speaking of the devil
`wb_add_comment(dims = "A1", comment = "foo")` this doesn't work! 

`Error: comment must be class wbComment or R6`

JanMarvin:
Yeah sorry, that was just pseudo code. It requires a `create_comment()` so this should work:

```R
wb_add_comment(dims = "A1", comment = create_comment("foo"))
```

And hence it should be fairly simple to hide `create_comment()` from the user, which raises the question, why wasn't it done in the first place? Same with `threads` and `sparklines` (oh, I just copied the idea from comment, but who came up with this stuff and why?). Why don't I "fix" or "improve" it? Add another if condition for something I do not use after all? I'd love to see a imessage/whatsapp like html bubble view for thread comments. That would be a cool feature for another helper package, even though I do not use MS365 at work and therefore won't ever have access to real life thread comments 😄

PR #755, opened 2023-08-23
provide basename2 for to long strings

JanMarvin:


JanMarvin:
Could you check if this works for you @olivroy ? It works in the Windows CI (R 4.1 and current)

olivroy:
Thanks for this!

It works. :0

PR #758, opened 2023-08-23
Export `wb_comment()`

olivroy:
Proof of concept as discussed in #754 

* This does not change `create_comment()` in any way. It will continue working as it did before.
* Change `visible` default to FALSE in `wb_comment()` to account for new Excel behaviour. (i.e. when I create a note in Excel, by default, it is not displayed)
* Use package options for default `author`.
* Allow a simple string to be passed to `comment` in `wb_add_comment()`. Most times, you just want to use the defaults.
* I did the testing by opening in Excel. Seemed to work fine.
* I did not add any deprecation message, but I can add it!

I will revamp examples and vignettes for that. So, I will be able to check if things are not working as expected.

I removed the default options to `wbComment` because it is not exported.
I completed the options list as some were missing from the list.


R/class-workbook-wrappers.R
```diff
@@ -3008,6 +3002,11 @@ wb_add_comment <- function(
   ) {
 
   assert_workbook(wb)
+
+  if (is.character(comment)) {
```
JanMarvin, line 3006: This should accept more cases and maybe only exclude class comment? Why e.g. shouldn't it be possible to pass a number or a fmt_txt() down or dunno even a vector? People get strange ideas ... eh, with people I mean myself :)

JanMarvin:
After my initial complains ... I can live with this and we can merge prior to 1.0, aka I have no objection. Please test with threaded comments too.

R/class-workbook-wrappers.R
olivroy: `is.character(fmt_txt("x")` is TRUE. So, this will be accepted. A number.. Not sure ! Maybe wait for user requests?

olivroy:
Should a deprecation warning be issued in `create_comment()` potentially only with the deprecation option?

olivroy:
Threaded comments do not seem to use `wbComment` objects, if I am not mistaken?

JanMarvin:
IIRC threaded comments use create_comment() internally. Even though it's a duplication I'd prefer to not yet deprecate `create_comment()` just a few days before the release. A `deprecating soon` option warning should be enough (similar to the camelCase one).

R/class-workbook-wrappers.R
olivroy: I tested that fmt_txt inputs work. And they show correctly in Excel. I added an example in `wb_add_comment()`

olivroy:
Sounds good. Everything is ready. Deprecation is only if the option is set, like other ones. I will finalize the cleanup of `create_comment()` in a follow-up PR. (and look closely at threaded comments) (tests etc. ) so that I can close if I make an unintended change.

FYI, I plan to 

* add more `@family` tags so that the pkgdown index doesn't fail too much and to set the ordering with the functions, rather than manually. It was a complex exercise to try to know every function, so that's why the process was not smooth.
* Remove the last `write_formula`/ `write_*` from examples and vignettes and test if the wb_add_formula alternative works in Excel. (test everything in excel obviously)

R/openxlsx2-package.R
```diff
@@ -73,6 +73,8 @@
 #' * `options("openxlsx2.orientation" = "portrait")` ## page orientation
 #' * `options("openxlsx2.sheet.default_name" = "Sheet")`
 #' * `options("openxlsx2.rightToLeft" = NULL)`
+#' * `options("openxlsx2.soon_deprecated" = FALSE)` ## warn if using camelCase
+#' * `options("openxlsx2.creator")` ## Default is Windows Username
```
olivroy, line 77:
```suggestion
#' * `options("openxlsx2.soon_deprecated" = FALSE)` ## If `TRUE` warn if  using functions or features that were recently deprecated.
#' * `options("openxlsx2.creator")` ## Default name used when creating `wb_workbook()` or adding comments
```

PR #760, opened 2023-08-23
Remove wbChartSheet from exports

olivroy:
That got me confused. `wb_hyperlink()` was never exported in the first place, but still had a .Rd page on the pkgdown website.

That solves it.

Removes `convertToExcelDate()` from function index. 
image

Edit: I took advantage of the check failures to slightly review the docs of `wb_add_chartsheet()`


JanMarvin:
Thanks, merged!

PR #764, opened 2023-08-24
Make `wb_get_named_regions()` a wrapper.

olivroy:
Since the only thing that `wb_get_named_regions()` did was to try to load it. 

Renamed `x` to `wb`. And left `x` there to make sure no code is broken...
Although if someone provided `wb_get_named_regions(x = file, TRUE)`, this would not fail gracefully.


JanMarvin:
Thanks, sorry that the message now shows me as co-author, I just resolved the merge conflict via github api and my sleepy head clicked to quick on the commit message body)

PR #765, opened 2023-08-24
Review threaded comment docs

olivroy:
So that a user who scan do not see `wb_get_person()` before. 

image

Thought I'd do that before taking a look at the internals

Edit: works in Excel! but interestingly. threaded comments are hidden in Excel 365, another justification for the change of default value in #758


JanMarvin:
Thanks for the heads up, could you add a visible argument to threaded comments? Guess that could be helpful.

olivroy:
Not sure how it will respected however? Possibly after 1.0? cause atm, I have to hover over it (with `create_comment()` as internal) even though the internal still uses create_comment() rn. (about to change it in another PR.

I observed that openxlsx2 add threaded comment with my system default.

JanMarvin:
Unless you have anything urgent, I'd like to close the development for 1.0 with this PR (and/or a second one with the visibilty argument). So that we can rest for the day and do a release tomorrow or early Saturday.

olivroy:
I have just 2 small left. I will be v quick!

JanMarvin:
Alright, no rush. I just have a few things going on over the weekend will be a bit unavailable early next week. There is no hard deadline :)

JanMarvin:
Ah, this can be merged? Looks fine, although I liked the previous user a bit more 😛

olivroy:
> can this be merged?
Yes

> I liked the previous user a bit more 

What do you mean? I don't mind tweaking!

My motivation is just that person can be many things if put out of context, could be wb creator author, but since this is very specific to threaded comments, I thought that for avoiding confusion to hide it a bit more.

JanMarvin:
You switched `openxlsx2` to something else "a person who likes ...", but that was not a serious comment. I don't really mind :)

JanMarvin:
Thank you, merged!

PR #766, opened 2023-08-24
internal: use `wb_comment()` in wb_add_thread + deprecations guide

olivroy:
THis aims to help users deal better with deprecations

* Change internal to `wb_comment()`.
* This will ensure that `options(openxlsx2.soon_deprecated)` works as intended and does not detect internal function use.
* An added benefit of #758 is also that comment are similar. (both accept text, although fmt_txt() is not supported in threads

just so you know, there will be some internal rework necessary to make comments "visible" but I would wait for a user request for that. https://github.com/JanMarvin/openxlsx2/pull/765#issuecomment-1692169367

I squeezed in an enhanced note on the deprecations that we can direct users to should problems arise.

No visual change before and after

Shows when hovering
image

Hidden when cursor is far

image
Edit: the openxlsx2_options name change is just that it was bugging me when navigating in files.. Saw only a single use in the pkg.

Edit2 : wb_add_thread doesn't error when not providing person_id. But it creates corrupt output (can't `wb$open()` after) in case you want to add protection for that.


R/class-workbook-wrappers.R
```diff
@@ -3118,7 +3118,8 @@ wb_add_thread <- function(
     reply     = FALSE,
     resolve   = FALSE
 ) {
-
+  # wb_add_thread now uses wb_comment internally. No change detected.
+  # means that add_thread does not look at visibility. (I think it's fine.)
```
JanMarvin, line 3122: yes, saw that too. visibility is set in wb$comment[[1]]$clientData which is nulled with threaded comments (that's what MS365 does. A comment is added for all spreadsheet software unaware of threaded comments. We can remove this comment in a cleanup commit.

olivroy: Good to know if ever need to revisist

JanMarvin:
Thank you so much <3

JanMarvin:
> Edit2 : wb_add_thread doesn't error when not providing person_id. But it creates corrupt output (can't wb$open() after) in case you want to add protection for that.

I wanted to capture this case here. If that's not working, we might have to revisit this. But it is again midnight in Germany and it's been a long week.

https://github.com/JanMarvin/openxlsx2/blob/7b65f82b7de8f792360d62b0e39bfc9006bc685c/R/class-workbook.R#L3820-L3824

PR #767, opened 2023-08-24
Improve functions index

olivroy:
This PR: If you agree with both this and #768, just merge #768. git problems again 

image

Also makes it easier to see that some things are not functions
image

Before:
image


JanMarvin:
you might want to look into a few git clients. For instance on Linux I prefer to use gitg (but that's just a personal choice) and its available via homebrew (I have a custom build in my github folder). The issue is, that you open new branches from dirty ones.

```R
git checkout main
git checkout -b dirty
git add somefile.R
git commit -m "some commit"
git checkout -b new # <- this carries the earlier commits from dirty over
```

In such a case, if you end up in new:
```R
git checkout main
git checkout -b new2
git cherry-pick ... # the commit you want to carry over (I like to use gitg for this, but every basic git client should do it)
```

JanMarvin:
all the mac clients are somewhat expensive iirc (not that I don't value software, but gitg does most of what I want and is open source). But github has a desktop client too., but this was confusing to me. The git client in Rstudio is somewhat limited, the one in Visual Studio code is okayish, but I don't really have experience with it.

olivroy:
I use these. They are so helpful https://usethis.r-lib.org/articles/pr-functions.html and they never make mistakes. Unfortunately, they are a bit slow.

along with gert too.

JanMarvin:
Ah I like to use git via command line, even though R is great, I don't want to use it for everything. But yeah, Jeroen is doing many cool things!

PR #768, opened 2023-08-24
Final reordering in pkgdown

olivroy:
close #767

* separate content and styling
* move down the more complex `chart_xml and drawing functions
* Add more content to first level as it tells a story of what order things should happen? I guess

image


JanMarvin:
Thanks! Good thinking (even though the changes are a bit hard to track for me :D)!

PR #769, opened 2023-08-24
Remove duplicated docs + remove internal functions docs.

olivroy:
This cleans up duplication.

You may just want to look at the .Rd diffs for this one as I used the powers of `roxygen2` to reduce duplication.

I realized I edited the same thing over and over. With this PR, it should be easier to manage.

Use `@inheritDotParams to help document write_xlsx.

Small tweaks to some function titles

Remove most docs for `write_*` functions.

Edit: I am officialy done. No more PRs from me for a while (hopefully)

Edit: I could have realized earlier I was wasting my time (and yours) tweaking documents for internal functions. So noob and sorry.

Some docs in the original functions could be improved.  but at least now, they are in a single place

image


JanMarvin:
We all live and learn. I once was joking that in `openxlsx` there are iirc 3 different functions for `int2col()`. At least one in R and one in c++. And I found this, will rewriting and moving code around. And what did I do? Create a fourth option and removed the others. Now that I've been developing `openxlsx2` for such a long time, I'm always afraid to do a similar thing, adding another `int2col()` function :-)

Thank you for the pull requests, thanks for sticking to it, I would have never done this cleanup, reordering and documentation stuff. Not in a hundred years. All the users have to salute you for all of this!

olivroy:
Thanks for the reviews (again)!

Hopefully people will like it. Anyway, for me, I will hate my life a little less when I am forced to use Excel. (and will not have to rely on SO answers about openxlsx to try to make it work)

A good doc >>>>> google!

PR #772, opened 2023-08-25
improve assert_class(). closes #770

### LINKED ISSUE #770: Suggestion: Better error message openxlsx2:::assert_workbook
## Background

`openxlsx2:::assert_workbook` is a function that checks the class of workbook-objects before performing operations on it:

```r
function (x) 
 assert_class(x, c("wbWorkbook", "R6"), all = TRUE)
```

I happened to write a `targets` pipeline missing a pipe from one workbook manipulation into the next, resulting in:

`Last error: wb must be class wbWorkbook or R6`

Which I find to be a little confusing.

## Suggestion

I think it would be nice to check for missing wb before asserting class, thus throwing a more precise error message, something along the way of:

```r
check <- function (wb) {
     if(hasArg(wb) == FALSE)
         stop("Workbook `wb` is missing")
     assert_class(wb, c("wbWorkbook", "R6"), all = TRUE)
 }
 
```

JanMarvin:


olivroy:
That's good. and about that, that's why I added `arg_nm` when I was developping `wb_dims()` in #702 

assert_class is not functioning entirely as expected.

```r
wb_comment(author = 12)
#> Error x must be class character.
```
It would be better if it was `author must be class character`.
It would be fixed by adding `arg_nm = "author"` in the assert code.
I wouldn't mind correcting that eventually but if that's an easy fix, feel free to look into it,

JanMarvin:
Thanks, yes I had a quick look at this, while preparing the release and figured something like this must have been the reason for `arg_nm`. I wasn't entirely sure if this would have side effects therefore it did not make the cut. All of the asserts were written by Jordan. I can have a look if there's some other way to simply tackle this. But first I'm gonna get some rest and afterwards write the release note :sleeping:

olivroy:
Indeed, it's not pressant. My next rounds of improvements will be regarding error messages. But possibly in a month or so!

Just wanted to highlight it here since it seemed related.

olivroy:
‘Object not found’ and ‘Missing argument’ errors now give a more accurate error context. Patch provided by Lionel Henry in [PR#18241](https://bugs.r-project.org/show_bug.cgi?id=18241).

I don't know how this looks like on R 4.3 vs R 4.2 (still using R 4.2, but maybe base R's patch is enough)

I just noticed that, and thought it was relevant, but possibly not. Please hide the comment if this is not relevant.

PR #779, opened 2023-08-31
internal: Rename files so that tests align a bit better

olivroy:
I did not redocument to avoid potential conflicts. 

They seem more clear this way.

maybe a file / test reorg / cleanup could be necessary eventually, but in my workflow I use,

`usethis::use_test()` when in R file to go edit the tests. this makes it a little easier.

Feel free to close, just wanted to try it out, (to me, it's just another way to further understand the interactions between functions in openxlsx2.


JanMarvin:
Thanks, but lets consider this sometime later. We have already had our share of file renames. This is something we can consider once most open pull requests are in.

JanMarvin:
Our tests are also a bit scrambled all over the place. We inherited this from `openxlsx`. But this is not an invitation to clean this up. Sometimes it simply makes a little more sense to test multiple functions at once. And currently I simply search with visual studio code for the functions and make my adjustments. That's why filenames don't really matter to me.

PR #780, opened 2023-08-31
assert_workbook() can now be used to load an empty workbook to return…

JanMarvin:
… sheet names without loading the sheets


olivroy:
Are you sure about this? This seems dangerous.
I would have added a hint message like:
Did you mean to load a file with `wb_load()` first?

Maybe just show this msg in interactive sessions.

JanMarvin:
No it's just a draft, but this would be what I mentioned in #775. Loading a workbook without sheets is helpful, if you have huge workbooks and don't want to load every workbook just to get the sheet names.

olivroy:
> Loading a workbook without sheets is helpful, if you have huge workbooks and don't want to load every workbook just to get the sheet names.
Ok, yes, this should be added to the `wb_load()` docs.

JanMarvin:
I just realized that we're suggesting that this works.
In our readme:
> Furthermore, there are several ways to read certain information of an openxml spreadsheet without having opened it in a spreadsheet software before, e.g. to get the contained sheet names or tables.

I doubt that this is true unless the workbook is loaded with `wb_load()`.

JanMarvin:
I close this in favor of #810

PR #752, opened 2023-08-22
Add possibility to interact with metadata

olivroy:
Fix #744, #745

The implementation is quite basic. There is now potential duplication of code, with `wb_add_creators()`, `wb_workbook()`, but it is not too bad. Feel free to tweak the implementation.

This seems to fail without worksheets, but I wouldn't know why you wouldn't want to have a worksheet in a workbook
Lmk what you think.
This now works https://github.com/JanMarvin/openxlsx2/issues/741#issuecomment-1682783730

```r
devtools::load_all("~/rrr-forks/openxlsx2")
#> ℹ Loading openxlsx2
wb1 <- wb_workbook(title = "thingy", subject = "A COOL SUBJECT")
wb1$add_worksheet()
wb1$open()
#> Warning in xl_open.default(x$clone()$save(temp_xlsx(macros = has_macros))$path,
#> : will not open file when not interactive
wb1$subject
#> [1] "A COOL SUBJECT"
wb1$set_properties(subject=  "A new subject")
wb1$subject
#> [1] "A new subject"
tmp_file <- temp_xlsx()
wb1$save(tmp_file)
wb2 <- wb_load(tmp_file)
wb2$subject
#> [1] "A new subject"
wb2$title
#> [1] "thingy"
```

edit: used the logic of wb$add_creator() mainly

Issue #46, opened 2022-01-22
modernize `self$print()` for R6 classes


JanMarvin:
The current output is way to detailed with many Linebreaks and information that could live an `verbose=TRUE` option.
Instead we could use `crayon` to color the sheets with tabColor etc.

With the current output I am sure that improvements can be made!

JanMarvin:
at least sheet names should be coma separated

PR #782, opened 2023-08-31
[properties] Rework setter/getter functions for workbook properties

JanMarvin:


JanMarvin:
@olivroy do you need wb$subject, wb$author fields etc? I had a look at the documentation and they allow various properties we currently might not have fields for [see here (xml at the bottom of the page)](https://sc34wg4.github.io/OOXMLSchemas/documentation/Part2/opc-coreProperties/opc-coreProperties_CT_CoreProperties.html).

My approach would be to keep everything in the XML string as is and modify only the few parts that we currently allow. With wb$get_properties() the user would have an option to get the values

JanMarvin:
I mean, we could fill the fields, but in this approach they would be useless

JanMarvin:
The core file can contain XML like this example (Ecma Office Open XML Part 1 - Fundamentals And Markup Language Reference, p. 5003) and we would lose all attributes.

```XML
<dc:creator>John Smith</dc:creator>
<property fmtid="{D5CDD505-2E9C-101B-9397-08002B2CF9AE}" pid="2" name="Editor">
<vt:lpwstr>John Smith</vt:lpwstr> </property>
```

olivroy:
Hi @JanMarvin,

> I mean, we could fill the fields, but in this approach they would be useless

I like filling the fields, it makes `wb_load()` more consistent with `wb_workbook()`. If we decide against filling them, it would be better to hide these fields from the class? (i.e. not show in the wb autocomplete.) and only let users access them with `wb$get_properties("author")`, but that's long. instead of `wb$author`

Possibly that `wb_workbook()` could allow dc:description too?

but in my PR, I shared what I thought would be a "nice to have" syntax, #752 

I was thinking that this metadata could be useful for printing #46 

In Excel 365, the following properties are used in core properties (for defining the content metadata) and show up for a spreadsheet user in the "Information" tab.
```
dc:title
dc:subject
dc:creator
cp:keywords
dc:description
```

I think all should be kept in `wb$core` and accessed with `wb$get_property()`, but maybe the ones that are arguments in `wb_workbook()` could be filled in.

Lmk what you think!

JanMarvin:
Yes, I have built upon your initial approach and I was impressed by the `modify_property()` function. Looks elegant and definitely not something I would have come up with.

But as stated I see issues and fear that we might lose some properties, which might or. might not be critical. There are many properties and unfortunately I have not yet seen a complete list of all possible properties (including custom properties or are they excluded and part of customXML and the link above contains the entire properties?) otherwise we could parse everything to and construct xml from a data frame similar to e.g. `read_xf()`/`write_xf()`.

What I want to avoid is something like the `openxlsx::stylesObject()` which looks great, but ignores (rare) special cases and converts valid styles into partial broken styles. If for whatever reason some workbook contains a strange file property, we should never remove it.

My approach would keep the xml untouched and only modify it in predefined cases. The currently not loaded fields `author`, `content` etc. can be removed.

Unfortunately this already looks a little more complex than I initially hoped.

olivroy:
I didn't want this to be too complicated for you. I am not familiar enough for all these XML safety considerations, so I will leave that to your judgement. 

Feel free to shelve if you don't think this is worth the effort.

olivroy:
Here is amother idea:

I think that removing the fields (description, creator, comments, etc. ) is a good start. (since they currently do not work, and are not useful. 

Use set_properties instead. the only consideration would be not to create invalid XML, but with tries we'd see how it is possible. 

The properties could be added to the print method too? 
A Workbook, title: <title>, categories: <categories> (only print properties that are set.

My use case would really be to load existing excel files (template ish), modify metadata and data, save to a new file


Btw #793 solved an issue I had, but couldn't quite figure out if I was doing something wrong, or it was an actual issue

JanMarvin:
I guess this PR already is aiming at what you want to do. I just got back yesterday and was very sleepy. Will work on this over the week. It's not forgotten, but sometimes good things take some time :)

JanMarvin:
can you with with this, @olivroy ?

R/class-workbook.R
```diff
@@ -5137,30 +5110,101 @@ wbWorkbook <- R6::R6Class(
...
+      }
+      if (!is.null(subject)) {
+        xml_properties["cp:category"] <- xml_node_create("cp:category", xml_children = category)
+      }
+      xml_properties["dcterms:created"] <- xml_node_create("dcterms:created",
+        xml_attributes = c(
+          `xsi:type` = "dcterms:W3CDTF"
+        ),
+        xml_children = format(as_POSIXct_utc(date_time_created), "%Y-%m-%dT%H:%M:%SZ")
+      )
```
JanMarvin, line 5164: move "namespace:node" text into variables

olivroy:
> can you with with this, @olivroy ?

Hi @JanMarvin, I will try with code once you finalize. Maybe `wb$properties` could be used as a shortcut to `wb$get_properties()`?

JanMarvin:
I don't really like shortcuts to internal functions. We already have a confusing amount of cases where it is not simple to guess what kind of input our R6 fields accept. Vectors or functions.
Other than that, if the CI is green this should be good to go.

olivroy (approved):
This looks good to me. I will test it in more depth in the coming days to see if edge cases have a problem. It's great that you were able to de-duplicate code and make wb_workbook and the `wbWorkbook` object safer at the same time.

Just because I don't see modifications to the load script. 

Are the properties picked up when loading?

JanMarvin:
Yes this is handled here. We load whatever is in `app/core.xml`:
https://github.com/JanMarvin/openxlsx2/blob/e9f24b9facc41a59a26840ae20cc0fda93baa5a2/R/wb_load.R#L244

The entire file is loaded into the string. You could test escaped characters `"you & me"` etc. I want to be on the safe side here, cause if we break things here, Excel might refuse to load a file or throw a warning. There are still some custom properties that are not handled here (the current approach is similar to `wb_set/get_core_properties()`. With custom properties you can add whatever you want (and I believe they are sometimes used for sharepoint access tokens, so it might be a good idea not to mess with these).

JanMarvin:
The only thing I'm not happy with in this pull request is the `creator`/`creators` mix. But it slipped into 1.0 and I'm not going to sacrifice my word to keep the API stable for a tiny little `s`.

olivroy:
Yeah! This provides a more consistent API for set/remove creators functions too! Possibly for a v2.0 set_creators/add_creators could be removed (or less advertised in favor of `wb_get/set_properties(creators = c("User 1", "User 2))` while leaving it exported.

olivroy:
FWIW, I don't see any usage of wb_add_creators, `remove_creators`, wb_set_creators()` on GitHub after a quick search.

JanMarvin:
We might begin a list for version 2.0 some time soon.

JanMarvin:
> FWIW, I don't see any usage of wb_add_creators, `remove_creators`, wb_set_creators()` on GitHub after a quick search.
> 
> Edit: a single use of `wb$add_creators`, in the eddible package but could open a future PR to change this for
> 
> ```r
> current_creator <- wb$get_properties()$creator
> wb$set_properties(creators = unique(c(current_creator, new_creator)))
> ```
> 
> but that's half-baked

Yeah, but we have time. I don't want to pressure people into changes that are part of a just released API.

PR #785, opened 2023-09-02
[tests] improve testfile_path() to download files 

JanMarvin:
Previously all files were downloaded prior to testing. Now the download is triggered by testfile_path(). This allows restoring parallel tests.

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -1001,9 +1001,13 @@ testfile_path <- function(x, replace = FALSE) {
 
   fl <- testthat::test_path("testfiles", x)
 
+  if (Sys.getenv("openxlsx2_testthat_fullrun") == "") {
+    if (isTRUE(as.logical(Sys.getenv("CI", "false")))) # on_ci()
+      return(testthat::skip("Skip on CI"))
+  }
+
   # try to download
   if (!file.exists(fl) || replace) {
-    # relies on libcurl and was optional in R < 4.2.0 on Windows
     out <- paste0(test_path, "/", x)
     url <- paste0("https://github.com/JanMarvin/openxlsx-data/raw/main/", x)
     try({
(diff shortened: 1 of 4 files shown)


JanMarvin:
Could you have a look @olivroy ? I remembered an earlier complaint from your side.

* Tests with external files should run only if the file was downloaded and downloads should only be attempted if online
* On the CI the tests with external files should only run on release and for coverage
* I didn't want to make things overly complicated, no md5 file hashes etc. If some test file is broken (e.g. partially downloaded, we can skip it)

tests/testthat/helper.R
```diff
@@ -988,72 +988,33 @@ expected_shared_strings <- function() {
...
+  fl <- testthat::test_path("testfiles", x)
 
-  # relies on libcurl and was optional in R < 4.2.0 on Windows
-  out <- paste0(test_path, "/", fls)
-  url <- paste0("https://github.com/JanMarvin/openxlsx-data/raw/main/", fls)
-  try({
-    download.file(url, destfile = out, quiet = TRUE, method = "libcurl")
-  })
+  if (Sys.getenv("openxlsx2_testthat_fullrun") == "") {
+    if (length(capture.output(try(testthat::skip_on_ci()))))
```
JanMarvin, line 1004: This is not working as intended. Github tests are skipped with "On CI" which is fine, but locally I had some issues that made me write this ...

olivroy:
Well, it's just that at first, the downloads were problematic..

As you said, it's probably fine as is.. 
I know that styler has a skip_if_parallel() that may be handy.

I will do more research on how other packages do this. the zip package creates temp dir for that too.

https://testthat.r-lib.org/articles/parallel.html the testthat vignette

olivroy:
I tripped on this again while working on #799 (solved by downloading the zip file from gh, unzip, then copy the contents manually to tests/testthat/testfiles

JanMarvin:
At least the code looks now a little less confusing

Issue #786, opened 2023-09-02
Removal of vignette images to reduce package size


JanMarvin:
Similar to the removal of most of the included example and test files, I want to remove the images from the vignettes and point people to https://github.com/JanMarvin/ox2-book for a detailed visual guide.

JanMarvin:
Closed in #1106

PR #789, opened 2023-09-06
Moving the conditional formatting vignette to an article

olivroy:
Addresses #786 

Since the vast majority of images is in the conditional formatting vignette, I think this is enough to close the issue.

Edit: source https://r-pkgs.org/website.html#vignettes-and-articles

https://r-pkgs.org/website.html#non-vignette-articles

> Content that involves a lot of figures, which cause your package to bump up against CRAN’s size constraints.
> The ggplot2 package [presents several FAQs as articles](https://ggplot2.tidyverse.org/articles/) for this reason.


JanMarvin:
Hi @olivroy , thanks for moving forward. I'm not really a huge fan of moving vignettes to articles. For this I prefer the quarto book. At least that creates html and pdf and will survive if pkgdown for whatever reason goes out of order. The entire package size thing is a bit random. I get that CRAN has size limits, but their binary file size varries a lot, which probably is related to nobody ever being able to reproduce their build infrastructure.

PS: I'm on a little vacation for the week and will be back on Monday.

olivroy:
enjoy your vacation ! 

I think a feature like r-lib/pkgdown#2265 sounds good for showing the imageless version on cran and only the one with images online.

JanMarvin:
The issue was closed in pkgdown. Do you want to carry this forward @olivroy ? I’m not sure that we will be able to get rid of the size note everywhere, but at least we could try.

If I understand the issue correctly, we can have an article version for the pkgdown website and a trimmed down vignette for the package.

olivroy:
Not sure here. I wonder how we plan to keep them in sync..

JanMarvin:
We could have a single Rmd file which includes graphics or urls depending on the article type or directory?

I would have preferred, if pkgdown would provide an option to include graphics only in the pkgdown documentation and not in the vignette. But maybe we can create something like this ourselves. After all it might be something like:

```r
# similar to testthat::on_cran()
if (!interactive() && !isTRUE(as.logical(Sys.getenv("NOT_CRAN", "false")))) {
  # so something like provide an url or so
} else {
  knitr::include_graphics("img/...")
}
```

But still not sure, how we can pass the images to pkgdown and exclude them from the tar file. Maybe knitr allows adding remote graphics.

olivroy:
I have an idea! will propose something later

PR #793, opened 2023-09-11
[remove_worksheet] update activeTab. closes #792

### LINKED ISSUE #792: Excel throws an error- "Removed Records: View from /xl/workbook.xml part (Workbook)"
I loaded a workbook from my local drive, delete a worksheet and save the workbook. When I tried to open the workbook by double-clicking the output file, Excel threw an error.

If I added a new worksheet, keeping the total number of worksheet same as the loaded workbook, no error was encountered.

The version of openxlsx2 is 1.0.
The version of R is 4.3.0.

Here is the reprex.

``` r
library(openxlsx2)
#> Warning: package 'openxlsx2' was built under R version 4.3.1

# Excel prompts error when the file is being opened
wb <- wb_load("test file.xlsx")
#> Error in wb_load("test file.xlsx"): File does not exist.
wb$get_sheet_names()
#> Error in eval(expr, envir, enclos): object 'wb' not found
wb$remove_worksheet(1)
#> Error in eval(expr, envir, enclos): object 'wb' not found
wb$get_sheet_names()
#> Error in eval(expr, envir, enclos): object 'wb' not found
wb_save(wb, "test_out.xlsx")
#> Error in eval(expr, envir, enclos): object 'wb' not found

# {no error
wb <- wb_load("test file.xlsx")
#> Error in wb_load("test file.xlsx"): File does not exist.
wb$get_sheet_names()
#> Error in eval(expr, envir, enclos): object 'wb' not found
wb$remove_worksheet(1)
#> Error in eval(expr, envir, enclos): object 'wb' not found
wb$get_sheet_names()
#> Error in eval(expr, envir, enclos): object 'wb' not found
wb$add_worksheet("new1")
#> Error in eval(expr, envir, enclos): object 'wb' not found
wb$get_sheet_names()
#> Error in eval(expr, envir, enclos): object 'wb' not found
wb_save(wb, "test_out.xlsx")
#> Error in eval(expr, envir, enclos): object 'wb' not found
# }
```

<sup>Created on 2023-09-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>
Excel's error message
[test file.xlsx](https://github.com/JanMarvin/openxlsx2/files/12571127/test.file.xlsx)

JanMarvin:
Could you please double check with your files that it works as expected @chowmun63?
```R 
# the id of the PR, should be 793 or something
remotes::install_github("JanMarvin/openxlsx2#793)
```

The issue was that the last tab was active, but removed and the activeTab id was never updated. Therefore Excel complained.

### DIFF
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -3357,24 +3357,6 @@ wbWorkbook <- R6::R6Class(
         self$tables$tab_act[sel] <- 0
       }
 
-      # ## drawing will always be the first relationship
-      # if (nSheets > 1) {
-      #   for (i in seq_len(nSheets - 1L)) {
-      #     # did this get updated from length of 3 to 2?
-      #     #self$worksheets_rels[[i]][1:2] <- genBaseSheetRels(i)
-      #     rel <- rbindlist(xml_attr(self$worksheets_rels[[i]], "Relationship"))
-      #     if (nrow(rel) && ncol(rel)) {
-      #       if (any(basename(rel$Type) == "drawing")) {
-      #         rel$Target[basename(rel$Type) == "drawing"] <- sprintf("../drawings/drawing%s.xml", i)
-      #       }
-      #       if (is.null(rel$TargetMode)) rel$TargetMode <- ""
-      #       self$worksheets_rels[[i]] <- df_to_xml("Relationship", rel[c("Id", "Type", "Target", "TargetMode")])
-      #     }
-      #   }
-      # } else {
-      #   self$worksheets_rels <- list()
-      # }
-
       ## remove sheet
       sn <- apply_reg_match0(self$workbook$sheets, pat = '(?<= name=")[^"]+')
       self$workbook$sheets <- self$workbook$sheets[!sn %in% sheet_names]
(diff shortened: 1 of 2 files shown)


chowmun63:
> Could you please double check with your files that it works as expected @chowmun63?
> 
> ```r
> # the id of the PR, should be 793 or something
> remotes::install_github("JanMarvin/openxlsx2#793)
> ```
> 
> The issue was that the last tab was active, but removed and the activeTab id was never updated. Therefore Excel complained.

You are right. It was my oversight on deleting the active worksheet. After activating the correct worksheet before deleting the inactive worksheet, I don't see any more error. Than you very much!

JanMarvin:
Thanks for the report. This is something that we do not handle entirely correct. The active tab is set in workbookView. We only support a single view, but according to the documentation multiple views are possible. Since I don't want to break all the views we keep them as is without any modifications. Unfortunately Excel error messages are very... unintuitive. It's more like "something is wrong". That's why something like this is easy to slip through. We fix things like this once they are reported.
btw great Bug report. it was easy to figure out the issue and fortunately a simple fix

PR #799, opened 2023-09-25
Review workbook properties + add `company` and `manager`

olivroy:
Sorry not to have caught this earlier. Follow-up to #782 

It is just that when I did a global search, I found 16 `datetime_created` and 11 `date_time_created`. This PR standardises everything to datetime_created. `wb_workbook()` uses datetime_created.

## Approach

global search and replace.


olivroy:
Should wb$get_properties() remove the dc: or cp: prefix from the print. i.e.

```r
# now
wb$get_properties()
         dc:creator            cp:keywords        dcterms:created      cp:lastModifiedBy 
                "me"                 "what" "2023-09-25T15:56:36Z"                 "me" 
# proposed
wb$get_properties()
         creator            keywords        datetime_created      modifiers
                "me"             "what" "2023-09-25T15:56:36Z"                 "me" 

```

I am looking into #46. So, this could be of interest.

In the class, you documented 

    #' @description Get properties of a workbook
    #' @param escape escape xml strings

but get_properties doesn't have an `escape` argument.

R/class-workbook-wrappers.R
```diff
@@ -2229,14 +2229,14 @@ wb_get_properties <- function(wb) {
...
-wb_set_properties <- function(wb, creators = NULL, title = NULL, subject = NULL, category = NULL, date_time_created = Sys.time(), modifiers = NULL, keywords = NULL, comments = NULL) {
+wb_set_properties <- function(wb, creators = NULL, title = NULL, subject = NULL, category = NULL, datetime_created = Sys.time(), modifiers = NULL, keywords = NULL, comments = NULL) {
   assert_workbook(wb)
   wb$clone()$set_properties(
     creators          = creators,
     title             = title,
     subject           = subject,
     category          = category,
-    date_time_created = date_time_created,
+    datetime_created = datetime_created,
```
JanMarvin, line 2239: please add an additional whitespace after the first `datetime_created`

R/class-workbook.R
```diff
@@ -212,7 +212,7 @@ wbWorkbook <- R6::R6Class(
         title             = title,
         subject           = subject,
         category          = category,
-        date_time_created = datetime_created,
+        datetime_created = datetime_created,
```
JanMarvin, line 215: here as well

JanMarvin:
Thanks, yes we can do that. Regarding the names, initially I wanted to keep things as they are, but now I think that we can replace the xml name with their initializing arguments and probably we should add manager and company to this as well.

Should be enough to add something like this:

```R
wb$app$Manager <- xml_node_create("Manager", xml_children = manager)
wb$app$Company <- xml_node_create("Company", xml_children = company)
```

JanMarvin:
Since you replied straight away, do you want to add this?

JanMarvin:
Oh get properties was from an earlier draft, I wanted to avoid something like `wb_workbook(author = "you & me")` break things. Special characters must be escaped and in the initial approach I wasn't sure if this works.

olivroy:
Oh, okay! And would you mind that I revert back from `creators` to `creator`. In `wb_workbook()`, `creator` is used. `creators` seems to be used only in `wb_*_creators()` functions, which become duplicates or outdated after the properties bit.

JanMarvin:
I just don't want to break our API. If I went to far, please revert. I got lost in `creator` and `creators` 🥲

olivroy:
Ok, so moved back creators to creator. And used the "R names" for properties. Planning to add the XML explanation to **Details** in `properties-wb`
Added the manager and company field. Since it is not core, I had to manually add them / get them

R/class-workbook.R
```diff
@@ -5162,7 +5185,7 @@ wbWorkbook <- R6::R6Class(
           core_subject = "",
           core_creator = "",
           core_keyword = "",
-          core_describ = "",
+          core_comments = "",
```
JanMarvin, line 5188: core_comment? I used always the same length for ... the looks ...

tests/testthat/test-loading_workbook.R
```diff
@@ -399,36 +399,38 @@ test_that("sheetView is not switched", {
 })
 
 test_that("Loading a workbook with property preserves it.", {
-  wb <- wb_workbook(title = "x", creator = "y", subject = "z", category = "aa", keywords = "ab", comments = "ac")$add_worksheet()
+  wb <- wb_workbook(title = "x", creator = "y", subject = "z", category = "aa", keywords = "ab", comments = "ac", manager = "ad")$add_worksheet()
   tmp <- temp_xlsx()
```
JanMarvin, line 403: Company for completeness?

JanMarvin:
looks good to me (y)

olivroy:
Comments addressed. Thanks. hopefully CI passes after the last remaining `creator`.
I just finished naming all XML core properties in details

JanMarvin:
Thanks, let me know when to merge. Iirc I named modifiers because of creators. Maybe we can switch this too, but I leave this for you to decide. We have to die one death ... (mostly because I'm a stubborn person).

olivroy:
> Thanks, let me know when to merge. Iirc I named modifiers because of creators. Maybe we can switch this too, but I leave this for you to decide. We have to die one death ... (mostly because I'm a stubborn person).

I resisted the urge of changing it. Now changing it with your blessing.

olivroy:
Ready to merge once CI passes. Thanks!

JanMarvin:
Thanks, merged!

PR #812, opened 2023-10-02
Tweaks to docs

olivroy:
To avoid duplication when documenting the R6 methods, now `sheet` and `dims`, and `...` doesn't have to be documented each time a method is added.

I also corrected a link and moved all examples out of the R6 docs to concentrate them in the wb wrappers, as it is more accessible there.


olivroy:
Good catch. I pushed another commit to address that.

olivroy:
@JanMarvin I squeezed in 
e2f93be to avoid loading testthat when using add_plot It isn't shown here. https://janmarvin.github.io/openxlsx2/reference/wb_add_plot.html?q=wb_add_plot#null but interactively, this loads testthat.

JanMarvin:
Thanks! Merged

PR #835, opened 2023-10-30
[write_data] make formula encoding more general

JanMarvin:
Realized that this is a more general move and covers a few more cases.

### DIFF
--- a/R/write.R
+++ b/R/write.R
@@ -220,20 +220,29 @@ write_data2 <- function(
   dc <- openxlsx2_type(data)
 
   # convert factor to character
-  if (any(dc == openxlsx2_celltype[["factor"]])) {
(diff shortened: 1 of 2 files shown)


R/write.R
```diff
@@ -220,20 +220,29 @@ write_data2 <- function(
...
+  )
+
   if (any(is_fml)) {
     fmls <- names(dc[is_fml])
     data[fmls] <- lapply(
       data[fmls],
-      function(val) xml_value(xml_node_create("fml", val, escapes = TRUE), "fml")
+      function(val) {
+        val <- replaceXMLEntities(val)
+        vapply(val, function(x) xml_value(xml_node_create("fml", x, escapes = TRUE), "fml"), "")
```
JanMarvin, line 244: might be replaced with `openxlsx2:::replace_legal_chars()`

PR #853, opened 2023-11-14
Document options

olivroy:
Follow-up to #848., where support for `maxWidth` and `minWidth` was added.

Related to #20

### DIFF
--- a/R/class-comment.R
+++ b/R/class-comment.R
@@ -118,7 +118,7 @@ wb_comment <- function(text = NULL,
                        width = 2,
                        height = 4) {
   # Code copied from the wbWorkbook
-  author <- author %||% Sys.getenv("USERNAME")
+  author <- author %||% Sys.getenv("USERNAME", unset = Sys.getenv("USER"))
   text <- text %||% ""
   assert_class(author, "character")
   assert_class(text, "character")
--- a/R/class-comment.R
+++ b/R/class-comment.R
@@ -118,7 +118,7 @@ wb_comment <- function(text = NULL,
                        width = 2,
                        height = 4) {
   # Code copied from the wbWorkbook
-  author <- author %||% Sys.getenv("USERNAME")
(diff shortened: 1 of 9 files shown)


R/class-comment.R
```diff
@@ -118,7 +118,7 @@ wb_comment <- function(text = NULL,
                        width = 2,
                        height = 4) {
   # Code copied from the wbWorkbook
-  author <- author %||% Sys.getenv("USERNAME")
+  author <- author %||% Sys.getenv("USERNAME") %||% Sys.getenv("USER")
```
olivroy, line 121: Oops, I just remember that `Sys.getenv()` returns "" when it can't find something. This will not work as intended.

JanMarvin: Can we grep for USER in the environment? Is it an issue with Windows? I tested with Linux and Mac and afaik they all had USERNAME

olivroy: For me (Windows) `Sys.getenv("USER")` returns `""` and `Sys.getenv("USERNAME")` works as expected. Let's drop "USER" then?

JanMarvin: 👍 you introduced it, you can remove it

JanMarvin: Darn! I lied. Sorry for the confusion :-(
My Mac shows
```R
> Sys.getenv("USERNAME")
[1] ""
> Sys.getenv("USER")
[1] "janmarvingarbuszus"
```

maybe something like:

```R
ifelse(Sys.getenv("USERNAME") == "", Sys.getenv("USER"), Sys.getenv("USERNAME"))
```

olivroy: Thanks for validating. I will reuse the `unset` argument of `Sys.getenv()` instead of ifelse.

JanMarvin:
Thanks, just one more thing, could you add `minWidth` to the code? Guess next to the `if widths > getOption(maxWidth)` part we might want something like: `if widths < getOption("minWidth")`.

JanMarvin:
Thanks, merged!

Issue #859, opened 2023-11-23
Features to document


JanMarvin:
For release 1.2
* various pivot tables (but PT need documenting in manual and vignette/book)
  * [ ] show_data_as #833
  * [ ] compact/outline #833
  * [ ] case names #836
  * [ ] choosing variables #840 / #858 (latter has example with picture)
  * [ ] item selection #842

* cloning column/row formatting #839

* [ ] improvements to copy cells #850/#851/#855 (needs documenting in manual, vignette and book)

* fixes
  * named region index #830
  * formula escaping
  * fixed wasm builds #849
  * restore min/max width options

Feel free to pick any or none @olivroy. Sorry, took me a moment. We probably might expand this issue for the future and it probably is no release stopper. Please let me know what you think.

JanMarvin:
I've extended the pivot table segment in [the book](https://janmarvin.github.io/ox2-book/chapters/openxlsx2_pivot_tables.html), since it requires a few images to explain what is supposed to happen. It's still a draft, but at least a little more documented.

JanMarvin:
Round two of documenting pivot tables is done: https://janmarvin.github.io/ox2-book/chapters/openxlsx2_pivot_tables.html

PR #867, opened 2023-12-04
Add reference to book

olivroy:
Address #859


JanMarvin (approved):
Thanks, looking good! Some day in the future I might add more details to the man page, but imho it’s the pictures that are most helpful in the chapter and CRAN does not want lots of pictures …

PR #873, opened 2023-12-12
Wb add style across

JanMarvin:
``` r
library(openxlsx2)

wb <- wb_workbook() %>%
  wb_add_worksheet() %>%
  wb_set_col_widths(cols = "A:B", width = 10) %>% 
  wb_add_fill(dims = "C3", color = wb_color("yellow"))%>% 
  wb_add_style_across(from = "C3", cols = "C:D", rows = 3:4)

if (interactive()) wb %>% wb_open()
```


JanMarvin:
Could you have a look at the API @olivroy ? I'm not entirely happy with `from`, but we've used it at other places and `dims` seemed misleading. Maybe you have a better idea.

olivroy:
@JanMarvin maybe `dims_ref`? 

If I understand correctly (I checked very quickly), it is that we take the style from a reference cell, and apply it to rows and cols?

`wb_copy_style_from(dims_ref = "C3", rows = 1:4)`? 

If not, I will take my time to think about it a bit more?

JanMarvin:
Thanks! Yes precisely, pick a style from a cell and apply it columns-wise/row-wise. I avoided new arguments, but it looks like a good idea.

JanMarvin:
I guess I rename and rework this function slightly. It might be helpful to rename this function to `wb_set_cell_style_across()` and make it behave similar to [`wb_set_cell_style()`](https://janmarvin.github.io/openxlsx2/reference/wb_cell_style.html). This would reduce confusing functions in `wb_add_style()`/`wb_add_style_across()` which behave differently and do different things.

olivroy:
@JanMarvin, maybe just add an argument to `wb_set_style()`. like the user should supply either `style` or `dims_ref`. or style could accept a cell
`wb_set_cell_style(dims = wb_dims(rows = 2:4, cols = 5:10),  copy_style_from = "A1")`

JanMarvin:
@olivroy I changed the `style` argument of `wb_set_cell_style()` to accept both a style id as before and a cell, and the newly added `wb_set_cell_style_across()` function behaves similarly. I don't think there is much demand for this kind of cell styles and now we have everything in one place and the functions behave similarly, so users who were using the old function should have no problem understanding the new one.

I appreciate your input as always, but in the end I decided against `dims_ref` and I'm not a big fan of the very long arguments you suggested in your last comment, even though I see the benefits they bring.

olivroy:
Thanks for your work! I think it is a great addition to use a cell as reference to style across the workbook!

PR #924, opened 2024-01-30
code lints

olivroy:
may improve (very) slightly performance. (Split from #922)


JanMarvin (approved):
Thanks! Feel free to merge

PR #936, opened 2024-02-12
Misc doc edits

olivroy:
Not yet done for the vignette, but thought I'd send these editorial changes.

I adapted what you put in `openxlsx2-package` to add to the class docs.

Corrected a link in README that takes us to an unused page as well.


JanMarvin (approved):
Thanks and good catch!

I was wondering, if we should add more `#' @inherits details wb_foo` in the class workbook documentation, because currently it is a long and rather useless dump of `#' @param foo foo` graveyard.  But I haven't looked into this and _\*sigh\*_  I do not really care enough

JanMarvin:
Oh alright, probably not worth the hassle. Though I somehow doubt that they will be ever solved.

PR #939, opened 2024-02-14
Change `num_fmt` to `numfmt`

olivroy:
This is incomplete here.

Just as a test here, but I wouldn't know how to adjust `standardize_case_names()` to make this actually work.

Let me know what you think

** No strong feelings, will likely end up closing, but giving a go anyway


JanMarvin:
I have at least two of these functions that standardize. One for camel case and one for color.  A third was in the `na_strings` one. For something like this I’d say a special one is needed for predefined sets. Standardize `_numfmt` into `_num_fmt`. Should be possible though must be carried out with care.

PR #942, opened 2024-02-14
Improve error message when no worksheet

olivroy:
Before this, I got 

```r
wb_workbook()$add_data(x = mtcars)
Erreur dans wb$is_chartsheet[[sheet]] : 
  tentative de sélection de moins d'un élément dans integerOneIndex

```

Also added an example of adding a URL hyperlink (I didn't know how to do this, so I documented it!) + added a NEWS bullet + re-rendered README since I forgot to do so in #936


README.md
```diff
@@ -54,7 +54,7 @@ other spreadsheet software are not supported). We offer two different
 variants how to work with `openxlsx2`.
 
 - The first one is to simply work with R objects. It is possible to read
-  ([`read_xlsx()`](https://janmarvin.github.io/openxlsx2/reference/read_xlsx.html))
+  ([`read_xlsx()`](https://janmarvin.github.io/openxlsx2/reference/wb_to_df.html)
```
JanMarvin, line 57: Have I caused a bad rebase?

R/write.R
```diff
@@ -743,7 +743,6 @@ write_data_table <- function(
 
   ## Input validating
   assert_workbook(wb)
-  if (missing(x)) stop("`x` is missing")
```
JanMarvin, line 746: This changes `write_data()` which is deprecated, but might still be used? But it's fine for me, too.

JanMarvin (approved):
Thanks, looking good!

R/write.R
olivroy: the evil me thinks this is fine, because if reported, we can point to the non-deprecated solution ;)!

README.md
olivroy: I forgot devtools::build_readme() + forgot the extra bracket.

R/write.R
JanMarvin: Hehe, that's one way to look at it

PR #943, opened 2024-02-14
Snake case conversion.

olivroy:
to avoid clutter!

### DIFF
--- a/R/write_xlsx.R
+++ b/R/write_xlsx.R
@@ -341,9 +341,10 @@ write_xlsx <- function(x, file, as_table = FALSE, ...) {
   }
 
   for (i in seq_len(nSheets)) {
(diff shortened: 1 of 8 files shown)


R/write_xlsx.R
```diff
@@ -358,6 +359,7 @@ write_xlsx <- function(x, file, as_table = FALSE, ...) {
         na.strings  = na.strings
       )
     } else {
+      # TODO add_data()?
```
JanMarvin, line 362: Ah, at first I was quite confused, but yes, we might look into this! The problem is mainly, I still don't really use this function, hence the overlooked fixes.

JanMarvin: But this can wait until the next release. I'm not losing sleep over this :)

JanMarvin (approved):
Thanks! I love your tiny PRs 😍

PR #956, opened 2024-02-21
[comment] export wb_get_comment()

JanMarvin:
Why does my code not run?

Oh, there already is a function called `wb_get_comment()`.

* [x] export `wb_get_thread()` too

Also ...
* `wb_comment()` could be an S3 class, no need for the R6
* I'd like to have something styled, but most ideas are to heavy. Either with third party packages like `cli` or javascript to create chart bubbles


JanMarvin:
Hi @olivroy , I picked `wb_get_comment()` to be in line with `wb_add_comment()`, but an additional `s` won’t do any harm I guess.

Could you elaborate the `dims = NULL` part? Right now all comments from a sheet are selected. And with `dims` this can be reduced to a selected region of a sheet. Why should we throw an error?

JanMarvin:
No worries, happens to me all the time 😄 But I’m not 100 percent sure I like to return everything as a data frame. It makes sense for the internal usage, but users probably would expect something nicer. Maybe I should just ditch the wbComment class and replace it with something easier to digest. I'd like some more styling, but haven’t made up my mind. I’ll think about it a bit more.

PR #958, opened 2024-02-24
[formulas] allow passing array formula vectors

JanMarvin:


R/write.R
```diff
@@ -308,8 +308,20 @@ write_data2 <- function(
...
+  is_array <- dc %in% c(
+    openxlsx2_celltype[["array_formula"]], openxlsx2_celltype[["cm_formula"]]
+  )
+
   if (!is.null(attr(data, "f_ref"))) {
     ref <- attr(data, "f_ref")
+  } else if (any(is_array)) {
+    ddims <- wb_dims(
+      x = data, from_row = startRow, from_col = startCol,
+      row_names = rowNames, col_names = colNames
```
JanMarvin, line 320: I haven't tested this to the limit. It works with the example ...

R/write.R
```diff
@@ -1122,10 +1142,21 @@ write_formula <- function(
...
+  if (is.null(dims)) {
+    dims <- wb_dims(start_row, start_col)
+  }
+
+  if (array || cm) {
+    if (length(dfx$X) == 1) {
       attr(dfx, "f_ref") <- dims
+    } else {
+      rowcols <- dims_to_rowcol(dims)
+      ddims   <- wb_dims(x = dfx$X, from_col = rowcols[[1]], from_row = as.integer(rowcols[[2]]))
```
JanMarvin, line 1156: @olivroy I was wondering, if we should add something like: `wb_dims(from_dim = "A1")`.

Or even `wb_dims(from_dims = "A1:D2")` and additional positioning arguments like `right`, `below`, `top`, `left`. I have quite a few real life cases where I have to position various data frames next to each other on a single sheet, where I could envision this as useful. Might even be a integer argument, so that `right = 2` would create a dims range two columns next to the original.

```R
dims <- wb_dims(x = mtcars)
dims
#> "A1:K33"

wb_dims(from_dims = dims, x = mtcars, right = 2)
#> "M1:W33"

wb_dims(from_dims = dims, x = mtcars, below = 2)
#> "A35:K67"
```

R/write.R
```diff
@@ -1069,6 +1081,14 @@ write_formula <- function(
   standardize_case_names(...)
 
   assert_class(x, "character")
+
+  # detect array formulas
+  if (any(substr(x, 1, 1) == "{")) {
+    pattern <- "^\\{(.*)\\}$"
+    x <- gsub(pattern, "\\1", x)
+    array <- TRUE
+  }
```
JanMarvin, line 1090: I took a confusing and ugly looking function and made it a little more confusing and ugly looking :)

olivroy: regex are always your friends when trying to make a function uglier!

R/write.R
JanMarvin: That would be lovely!

PR #960, opened 2024-02-29
Add from_dims to wb_dims

olivroy:
Addresses part of https://github.com/JanMarvin/openxlsx2/pull/958#discussion_r1501437063

I made `from_dims` independant from `from_col` / `from_row`, but didn't really document it. Hopefully the error should be clear enough?

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -518,7 +518,8 @@ wb_dims <- function(..., select = NULL) {
 
   # nams cannot be NULL now
   nams <- names(args) %||% rep("", len)
-  valid_arg_nams <- c("x", "rows", "cols", "from_row", "from_col", "from_dims", "row_names", "col_names")
+  valid_arg_nams <- c("x", "rows", "cols", "from_row", "from_col", "from_dims", "row_names", "col_names",
+                      "left", "right", "above", "below")
   any_args_named <- any(nzchar(nams))
   # unused, but can be used, if we need to check if any, but not all
   # Check if valid args were provided if any argument is named.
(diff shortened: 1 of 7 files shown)


JanMarvin:
Thanks for getting the ball rolling. I was thinking about something like this:

``` r
library(openxlsx2)
dims <- wb_dims(from_dims = "A1", x = mtcars)

wb_dims(from_dims = dims, x = mtcars)
#> Error in wb_dims(from_dims = dims, x = mtcars): `from_col` / `from_row` should have length 1. and be positive.
```

R/utils.R
```diff
@@ -587,10 +589,23 @@ wb_dims <- function(..., select = NULL) {
...
+  # handle from_dims
+
+  if (!is.null(args$from_dims)) {
+    if (!is.null(args$from_col) || !is.null(args$from_row)) {
+      stop("Can't handle `from_row` and `from_col` if `from_dims` is supplied.")
+    }
+    # transform to
+    from_row_and_col <- dims_to_rowcol(args$from_dims, as_integer = TRUE)
+    args$from_col <- from_row_and_col[[1]]
+    args$from_row <- from_row_and_col[[2]]
```
JanMarvin, line 601: Probably just like this:
```r
args$from_col <- from_row_and_col[[1]][length(from_row_and_col[[1]])]
args$from_row <- from_row_and_col[[2]][length(from_row_and_col[[2]])]
```

JanMarvin:
My intention is to be able to be able to further simplify the placement with `wb_dims(from_dims)`. So that if I want to write two datasets next to each other, I could simply use the dims from the first and say, write this to the top, left, bottom, right of this dims range. Picking just one dims and let the `wb_dims()` magic do the rest.

Does that make sense?

R/utils.R
```diff
@@ -587,10 +590,64 @@ wb_dims <- function(..., select = NULL) {
...
+
+    if (!is.null(left)) {
+      fcol <- min(from_col) - left - ncol(args$x) + 1L
+      frow <- min(from_row)
+    } else if (!is.null(right)) {
+      fcol <- max(from_col) + right
+      frow <- min(from_row)
+    } else if (!is.null(above)) {
+      fcol <- min(from_col)
+      frow <- min(from_row) - above - nrow(args$x)
```
JanMarvin, line 622: this might require a column name correction

olivroy: Oh no, this is hellish a bit. Took me a while to figure it properly without these.

olivroy:
I see what you mean. This is more complex than I thought.  I can test it a bit more thoroughly once you have merged. Let's call it very experimental for now.

JanMarvin:
Should work now:

``` r
library(openxlsx2)
dims <- wb_dims(from_dims = "B4", right = 4)
wb <- wb_workbook()$add_worksheet()$add_data(dims = dims, x = mtcars, row_names = TRUE)

wb$add_fill(dims = wb_dims(from_dims = dims, x = mtcars, row_names = TRUE, cols = "mpg"), color = wb_color("yellow"))

if (interactive()) wb$open()
```

R/utils.R
JanMarvin: Hehe, yeah, you've created a complex monster with `wb_dims()`. I have implemented something, maybe have a look if you have time. I probably wont have much time to work on this the upcoming week. Was thinking about corner cases, because ... why not `wb_dims(from_dims = "A2:B3", below = 4, right = 5)`. But we can wait until anyone asks for it. Probably going with what we have and documenting it a bit should be enough for this PR.

olivroy:
I guess this is good to go. I will try to test more thoroughly after

JanMarvin:
Thanks, merged

PR #967, opened 2024-03-12
[option] openxlsx2.nastrings option as discussed in #964

JanMarvin:
The only thing that is not possible, writing `na.strings = NULL`, because this will fall through. Writing `na.strings = NULL` suppresses writing anything, but you could write `na.strings = ""`. Does this work for you @olivroy?

### DIFF
--- a/R/write.R
+++ b/R/write.R
@@ -783,6 +783,10 @@ write_data_table <- function(
     return(wb)
   }
 
+  # overwrite na.strings with whatever is in the option
+  na.strings <- getOption("openxlsx2.nastrings") %||% na.strings
+
+
   if (data_table && nrow(x) < 1) {
     warning("Found data table with zero rows, adding one.",
             " Modify na with na.strings")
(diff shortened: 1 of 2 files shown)


R/write.R
```diff
@@ -783,6 +783,10 @@ write_data_table <- function(
     return(wb)
   }
 
+  # overwrite na.strings with whatever is in the option
+  na.strings <- getOption("openxlsx2.nastrings") %||% na.strings
```
olivroy, line 787: I think it would be better to peek at the option only if something more explicit was not provided?

```r
 dd <- mtcars
  dd[dd < 3] <- NA
options(openxlsx2.nastrings = "")
wb <- wb_workbook()
wb$add_worksheet()
wb$add_data(x = dd) # na.strings = ""
wb$add_data(x = dd, dims = "AA1", na.strings = "NA") # na.strings = "NA"
```

But I didn't submit a PR, because I didn't really know how to deal with this, so I would understand if you don't want to put efforts on this!
wb$add_data

JanMarvin: Not sure how you want to make this simpler. This option will only trigger if something is set. Otherwise the default is picked. You could consider it the secret olivroy option:)

JanMarvin:
Closing in favor of #968

PR #968, opened 2024-03-13
Add openxlsx2.na.strings option

olivroy:
I was thinking about something more like this.

I strongly feel about not having an option overriding a parameter passed (other than the default)

If option is set and na.string is  not supplied: use the option
If option is set and na.string is supplied: use na.strings
If option is not set and na.string is not use default.

Amends to #473 and addresses #20

### DIFF
--- a/R/write.R
+++ b/R/write.R
@@ -10,7 +10,8 @@
 #' @param colNames has colNames (only in update_cell)
 #' @param removeCellStyle remove the cell style (only in update_cell)
 #' @param na.strings Value used for replacing `NA` values from `x`. Default
-#'   `na_strings()` uses the special `#N/A` value within the workbook.#' @keywords internal
+#'   `na_strings()` uses the special `#N/A` value within the workbook.
+#' @keywords internal
 #' @noRd
 inner_update <- function(
     wb,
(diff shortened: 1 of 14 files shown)


R/write.R
```diff
@@ -782,6 +782,12 @@ write_data_table <- function(
   if (is.null(x)) {
     return(wb)
   }
+  # overwrite na.strings if nothing was provided
+  # with whatever is in the option if not set to default
+  if (identical(as.character(na.strings), "na_string") &&
```
JanMarvin, line 787: I don’t recall at the moment if we could check for inherits waiver.

JanMarvin:
I see why you want it this way. Since I have no preference, I’m equally happy with this PR!

R/write.R
olivroy: there is `is_na_strings()`, but I am somewhat scared of it.

R/write.R
```diff
@@ -782,6 +785,12 @@ write_data_table <- function(
...
+      !is.null(getOption("openxlsx2.na.strings"))) {
```
olivroy, line 791:
```suggestion
  if (is_na_strings(na.strings) && !is.null(getOption("openxlsx2.na.strings"))) {

```

R/write.R
```diff
@@ -782,6 +785,12 @@ write_data_table <- function(
   if (is.null(x)) {
     return(wb)
   }
+  # overwrite na.strings if nothing was provided
+  # with whatever is in the option if not set to default
+  if (is_na_strings(na.strings) && !is.null(getOption("openxlsx2.na.strings"))) {
+
```
olivroy, line 791:
```suggestion
```

JanMarvin:
Looks good to me! Thanks

PR #970, opened 2024-03-17
Fix docs

### LINKED ISSUE #969: [docs] CRAN warning regarding html
https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-gcc/openxlsx2-00check.html

```
checking HTML version of manual ... [8s/17s] NOTE
Found the following HTML validation problems:
convert_date.html:55:1 (convert_date.Rd:23): Warning: trimming empty <dt>
```

olivroy:
fix #969 

as.Date generic doesn't have any params, which resulted in empty \describe{}

This probably fixes it.

### DIFF
--- a/R/dates.R
+++ b/R/dates.R
@@ -8,7 +8,7 @@
 #' @seealso [wb_add_data()]
 #' @param x A vector of integers
 #' @param origin date. Default value is for Windows Excel 2010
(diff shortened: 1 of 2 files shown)


JanMarvin (approved):
Thanks! I would have rewritten it to include a hardcoded string. Good catch!

Issue #972, opened 2024-03-20
Add formulas to row


vladimirovnikita:
Hello!
Please tell me how to add formulas to a row?
The code below does not do what I expect and inserts formulas into column "A"

`wb <- wb_workbook()

wb$add_worksheet()

wb$add_data(x=data.frame(
                       `A`=c(1,2),
                       `B`=c(3,4),
                       `C`=c(5,6)),
            col_names = FALSE)

wb$add_formula(x = c("A1*A2","B1*B2","C1*C2"), 
               dims = "A3:C3")  

wb_save(wb, file = "test.xlsx")`

JanMarvin:
Hi @vladimirovnikita ,

currently positioning formulas with `dims` is not yet possible. Though I will open a pull request later today that I have been working on lately, that might change this.

For the time being you can use the following. This constructs a data frame with a formula class that contains the formula that you want to write. A second approach would be to loop over the cells with formulas that you want to write and place them with `dims`.

``` r
library(openxlsx2)
packageVersion("openxlsx2")
#> [1] '1.4'

wb <- wb_workbook()

wb$add_worksheet()

wb$add_data(x=data.frame(
  `A`=c(1,2),
  `B`=c(3,4),
  `C`=c(5,6)),
  col_names = FALSE)

# formula data frame that can be used to write formulas row wise
fml <- data.frame(
  sumA = "A1*A2",
  sumB = "B1*B2",
  sumC = "C1*C2"
)
class(fml$sumA) <- c("formula", "character")
class(fml$sumB) <- c("formula", "character")
class(fml$sumC) <- c("formula", "character")

wb$add_data(x = fml, dims = "A3:C3", col_names = FALSE)

if (interactive()) wb$open()
```

Issue #469, opened 2022-12-10
[write] be strict with dims area


JanMarvin:
Currently when providing dims to write data, we should provide an option to check if the area we are writing to is equal to dims and otherwise we should inform the user, maybe even abort if they don't match. Something like this:

* user writes 4 x 4 matrix into single cell. error
  * user writes 4 x 4 matrix into 2 x 4 cell. optional warning
  * user writes 4 x 4 matrix into 4 x 2 cell. optional warning
* user writes 1 cell to 4 x 4 matrix. optional warning

In a second attempt we should respect the user requested cells. If dims "A1:D1;B2:E2" is requested, only these cells should be filled.

This should include #37.

PR #974, opened 2024-03-20
[write] dims impact the direction vectors are written

JanMarvin:
This introduces a change in behavior when writing vectors. Previously, we simply ignored whichever direction `dims` requested. We just selected the top left corner and started writing from there. This PR changes that. We now use the direction that `dims` suggests:

* `A1` (the default) or `A1:A2` - vertical
* `A1:B1` - horizontal

and we transpose the vector so that it can be written horizontally. This affects vectors, formulas (#972) and hyperlinks.

The break occurs if a user has previously transposed a vector manually and also specified the direction via `dims`. In this case, the data is transposed twice and the output looks different with the pull request, see the first example.

``` r
library(openxlsx2)

# write vectors column or row wise
  add_data(x = 1:2, dims = "A1:B1")$
  add_data(x = t(1:2), dims = "D1:D2", col_names = FALSE)$
  add_data(x = 1:2, dims = "A3:A4")$
  add_data(x = t(1:2), dims = "D3:E3", col_names = FALSE)$
  add_fill(dims = "A1:B2", color = wb_color(theme = 4))$
  add_fill(dims = "A3:B4", color = wb_color(theme = 5))$
  add_fill(dims = "D1:E2", color = wb_color(theme = 6))$
  add_fill(dims = "D3:E4", color = wb_color(theme = 7))

if (interactive()) wb$open()


## sum, sum as array and sum as cm
wb <- wb_workbook()$
  add_worksheet()$
  add_data(x = head(cars))$
  add_formula(x = c("SUM(A2:A7)", "SUM(B2:B7)"), dims = "A9:B9")$
  add_formula(x = c("{SUM(A2:A7)}", "{SUM(B2:B7)}"), dims = "A10:B10")$
  add_formula(x = c("{SUM(A2:A7)}", "{SUM(B2:B7)}"), dims = "A11:B11", cm = TRUE)
#> Warning in write_data2(wb = wb, sheet = sheet, data = x, name = name, colNames
#> = colNames, : modifications with cm formulas are experimental. use at own risk

if (interactive()) wb$open()


## example from wb_add_formula()
wb <- wb_workbook()$add_worksheet()
wb$add_data(dims = wb_dims(rows = 1, cols = 1:3), x = c(4, 5, 8))

# calculate the sum of elements.
wb$add_formula(dims = "D1", x = "SUM(A1:C1)")

# array formula with result spanning over multiple cells
mm <- matrix(1:4, 2, 2)

wb$add_worksheet()$
  add_data(x = mm, dims = "A1:B2", col_names = FALSE)$
  add_data(x = mm, dims = "A4:B5", col_names = FALSE)$
  add_formula(x = "MMULT(A1:B2, A4:B5)", dims = "A7:B8", array = TRUE)

if (interactive()) wb$open()
```

Could you take a look, @olivroy, if you are following the intent of this PR? Unfortunately, when I wrote `write_data()`, `write_data2()` and `write_data_table()`, I didn't force a conversion to a data frame and kept building on the confused `openxlsx` logic, so the code looks terrible and is spread across three functions that should be rewritten in a single clean function.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,11 @@
 
 * Export `wb_add_ignore_error()`. [955](https://github.com/JanMarvin/openxlsx2/pull/955)
 
+## Breaking changes
+
+* The direction vectors are written is now controlled via `dims`. Previously it was required to transpose a vector to write it horizontally: `wb_add_data(x = t(letters), col_names = FALSE)`. Now the direction is defined by `dims`. The default is still to write vectors vertically, but for a horizontal vector it is possible to write `wb_add_data(x = letters, dims = "A1:Z1")`. This change impacts vectors, hyperlinks and formulas and basically everything that is not a two dimensional `x` object.
+This is only breaking, if previous code made use of the transposed trick above and already defined a correct `dims`. Since the `wb_add_data()` function is agnostic of the direction of `x`, this `wb_add_data(x = t(letters), dims = "A1:Z1", col_names = FALSE)` would write the `letters` into cells `A1:A26`, because of a round trip through transpose.
+
 
 ***************************************************************************
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,6 +19,12 @@
 * The direction vectors are written is now controlled via `dims`. Previously it was required to transpose a vector to write it horizontally: `wb_add_data(x = t(letters), col_names = FALSE)`. Now the direction is defined by `dims`. The default is still to write vectors vertically, but for a horizontal vector it is possible to write `wb_add_data(x = letters, dims = "A1:Z1")`. This change impacts vectors, hyperlinks and formulas and basically everything that is not a two dimensional `x` object.
 This is only breaking, if previous code made use of the transposed trick above and already defined a correct `dims`. Since the `wb_add_data()` function is agnostic of the direction of `x`, this `wb_add_data(x = t(letters), dims = "A1:Z1", col_names = FALSE)` would write the `letters` into cells `A1:A26`, because of a round trip through transpose.
 
+```r
+# before (workaround needed)
+wb$add_data(dims = wb_dims(rows = 1, cols = 1:3), x = t(c(4, 5, 8)), col_names = FALSE)
+# now (listens to dims)
+ wb$add_data(dims = wb_dims(rows = 1, cols = 1:3), x = c(4, 5, 8))
+```
 
 ***************************************************************************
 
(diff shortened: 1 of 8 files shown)


olivroy:
I think I get the idea. 

2 questions / suggestions:

1. Does setting the option to TRUE yield a lot of warning in openxlsx2 code base?
2. Maybe worth advertising this option in R/openxlsx2-package.R in the depreated topic. + in NEWS

NEWS.md
```diff
@@ -14,6 +14,11 @@
 
 * Export `wb_add_ignore_error()`. [955](https://github.com/JanMarvin/openxlsx2/pull/955)
 
+## Breaking changes
+
+* The direction vectors are written is now controlled via `dims`. Previously it was required to transpose a vector to write it horizontally: `wb_add_data(x = t(letters), col_names = FALSE)`. Now the direction is defined by `dims`. The default is still to write vectors vertically, but for a horizontal vector it is possible to write `wb_add_data(x = letters, dims = "A1:Z1")`. This change impacts vectors, hyperlinks and formulas and basically everything that is not a two dimensional `x` object.
+This is only breaking, if previous code made use of the transposed trick above and already defined a correct `dims`. Since the `wb_add_data()` function is agnostic of the direction of `x`, this `wb_add_data(x = t(letters), dims = "A1:Z1", col_names = FALSE)` would write the `letters` into cells `A1:A26`, because of a round trip through transpose.
+
```
olivroy, line 21:
````suggestion

```r
# before (workaround needed)
wb$add_data(dims = wb_dims(rows = 1, cols = 1:3), x = t(c(4, 5, 8)), col_names = FALSE)
# now (listens to dims
 wb$add_data(dims = wb_dims(rows = 1, cols = 1:3), x = c(4, 5, 8))
```
````

I like including code examples, sometimes easier to understand than words. Example: in [dplyr 1.1 news](https://dplyr.tidyverse.org/news/index.html#new-features-1-1-0)

JanMarvin:
Thanks! Regard 1) I haven’t actually tried that. Good point. 2) Same, should give it a try and will update NEWS.

btw how do you create these changes ready for commits?

olivroy:
> btw how do you create these changes ready for commits?

It is code suggestions, pretty handy! https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-comments-to-a-pull-request

JanMarvin:
Thanks!

JanMarvin:
Thanks to your suggestion, I've fixed the logic and added tests for it. The PR is related to #469, though at the moment there is at least one issue I see and why I haven't documented it further: `write_xlsx()` does not use `dims` and we'd have to fix this, otherwise every call with `write_xlsx()` and the `openxlsx2.warn_if_dims_dont_fit` option will throw a warning ... aaand I'm currently not in the mood to figure this out.

PR #983, opened 2024-03-29
[WIP] add MIP support. closes #979

### LINKED ISSUE #979: Handling of MIP Sensitivity Labels
As discussed in an `openxlsx` issue (https://github.com/ycphs/openxlsx/issues/461#issuecomment-1901162268) it is possible to load and apply sensitivity labels with `openxlsx2`.

There are things to consider
1) They are only visible in corporate environments that have MIP (Microsoft information protection) enabled (https://learn.microsoft.com/en-us/information-protection/develop/concept-mip-metadata) and there is no way for us to test this without such an environment that at the moment is not available.

2) The entire thing is just a custom xml string attached to the package (the security aspect is therefore similar to printing "Confidential" on each page of a document, just that this is at least visible. Outside of MIP environments the sensitivity label is no longer visible. So if someone shares a file marked as confidential with somebody in the outside world, this other person would not be aware of any confidentiality requirements, when opening the file and you can simply remove the sensitivity label by removing the xml string from the file ...).
Each MIP environment can provide custom roles for access. These roles have a custom id known only to the MIP environment and this is stored with the xml string. (Think of it as a corporate id string added to the "Confidential" stamp. For the sensitivity to apply, the stamp is not only "Confidential", but "Confidential for Corporation"). This id is required to write a valid sensitivity label. The id and the possible sensitivity levels could be obtained using the [power shell command](https://learn.microsoft.com/en-us/powershell/module/exchange/get-label?view=exchange-ps) `Get-Label`, but access might be restricted. 
Therefore extracting the `custom.xml` string from a template file should be the way to go. I somehow doubt that there is any security risk involved if the string is public, worst thing, somebody could stamp documents with this (similar to printing "Confidential for Corporation" on non confidential pages not from the Corporation). Still it might be a good idea to store the string in an option, so that if the user shares code, it will be obvious that the MIP string should not be passed along to everyone everywhere, but first and foremost, so that the user wont have to remember passing the string to every function.

3) We have to make sure that we do not duplicate these strings (Excel might complain if we do). Even if we do not yet provide a way to create or alter `custom.xml`, our function should not remove existing `custom.xml` strings. Not sure how to handle a situation, where the user might want to raise or lower the sensitivity of a document.

A few additional links:
* https://alberthoitingh.com/2021/12/01/new-metadata-model-mip/
* https://andrewwarland.wordpress.com/2022/04/04/sensitivity-labels-in-microsoft-365/
* https://learn.microsoft.com/en-us/purview/sensitivity-labels?view=o365-worldwide

JanMarvin:
TODO list

* [ ] feedback: does it work?
* [ ] move code from wrapper function to workbook function
* [ ] cleanups?

``` r
library(openxlsx2)
wb_w_mips <- wb_load("~/Downloads/SampleSensitivityIssue.xlsx")
#> Warning: Found unknown folders in the input file:
#> webextensions
#> 
#> These folders are not yet processed by openxlsx2, but depending on what you want to do, this may not be fatal.
#> Most likely a file with these folders has not yet been detected. If you want to contribute to the development of the package, maybe just to silence this warning, please open an issue about unknown content folders and, if possible, provide a file via our Github or by mail.

mips_xml <- wb_get_mips(wb_w_mips)
#> Found MIPS section: General

wb <- wb_workbook() |> 
  wb_add_worksheet() |> 
  wb_set_properties(
    custom = list(
      Software    = "openxlsx2",
      Version     = 1.5, 
      ReleaseDate = as.Date("2024-03-26"),
      CRAN        = TRUE,
      DEV         = FALSE
    )
  )

wb <- wb |>
  wb_add_mips(xml = mips_xml)

updated <- wb_get_mips(wb)
#> Found MIPS section: General

wb <- wb |>
  wb_add_mips(xml = mips_xml)
#> File has duplicated custom section

updated2 <- wb_get_mips(wb)
#> Found MIPS section: General

# expect difference, because of pid differences
all.equal(updated, updated2)
#> [1] TRUE

options("openxlsx2.mips_xml_string" = mips_xml)

wb <- wb |>
  wb_add_mips()

updated3 <- wb_get_mips(wb)
#> Found MIPS section: General

all.equal(updated2, updated3)
#> [1] TRUE

if (interactive()) wb$open()
```

Maybe you can have a look at this PR @retodomax

It's still WIP, but this should be a start. I only have this single file to play around with (it's from this issue: https://github.com/OfficeDev/office-js/issues/2459) and as stated no way to test if it works. My only test is that the output opens in MS365, but there is no MIP banner for me.

[EDIT] updated code snippet above


JanMarvin:
I've improved `wb_set_properties()` to accept custom file properties. Maybe you're interested @olivroy. Find an example above how to write custom text, numeric, date and bool properties.

The sensitivity label applied via MIP (the Microsoft Information Property thingy, see #979) are 7(?) custom text properties. Since there is probably no way for us to write this from scratch, I choose to read the xml string and apply this where required.

JanMarvin:
Thanks for the feedback @retodomax ! Odd that it doesn't work with the options. There is no reason I can think of, why it shouldn't. But maybe I forgot something, I'll have a look later.

Could you have a look which of the sections in mips_xlsx looks like a name? I assume there should be something that looks human readable like `General` or `Confidential` etc? But maybe it is just numbers in your organization.

```R
mips_xml |> as_xml()
```

retodomax:
Unfortunately, there is nothing like `General` or `Confidential`...

```
<property fmtid="{D5CDD505-...AE}" pid="2" name="MSIP_Label_1c79e38c-78...05_Enabled">
 <vt:lpwstr>true</vt:lpwstr>
</property>
<property fmtid="{D5CDD505-...AE}" pid="3" name="MSIP_Label_1c79e38c-78...05_SetDate">
 <vt:lpwstr>2024-04-02T06:00:28Z</vt:lpwstr>
</property>
<property fmtid="{D5CDD505-...AE}" pid="4" name="MSIP_Label_1c79e38c-78...05_Method">
 <vt:lpwstr>Privileged</vt:lpwstr>
</property>
<property fmtid="{D5CDD505-...AE}" pid="5" name="MSIP_Label_1c79e38c-78...05_Name">
 <vt:lpwstr>1c79e38c-78...05</vt:lpwstr>
</property>
```

and it continues with `SiteId`, `ActionId` and `ContentBits`. I don't know from where our internal Excel gets the information that this is a `Public` file... probably there is somewhere another file which links the numbers and the labels.

JanMarvin:
Thanks! Guess, as long as it works it does not have to bother us, but in this case I can disable the message. It should be used so that somebody how has multiple MIPS sections available, can differentiate them, but I doubt that anybody understands these digits.

It looks like they match your organizations MIPS string (from your example above, I've added `**` to highlight the presumably matching sections). Maybe a public document has no need for a human readable name.

```XML
<property fmtid="{D5CDD505-...AE}" pid="5" name="MSIP_Label_**1c79e38c-78...05**_Name">
 <vt:lpwstr>**1c79e38c-78...05**</vt:lpwstr>
</property>
```

Two additional observations:
* Out of curiosity, can you run `Get-Label` in Powershell? It's unlikely that I'll add a function to create these MIPS xml strings from scratch, but that might give another hint regarding the possible sensitivity levels.
* Do you see the timestamp of the sensitivity label in Excel? Should we update this timestamp when writing the file? Aka if you modify it, does it matter? (you can simply `dput()` the string into a notepad, change it to some arbitrary date and assign the string as `mips_xml` in the example code above).

retodomax:
Hi @JanMarvin 

1) Returns an Error "Get-Label : The term 'Get-Label' is not recognized as the name ..."
2)  I don't see any timestamp associated with sensitivity label. I don't notice any difference if I change it manually.

JanMarvin:
Hi @retodomax , thanks for the testing! Guess it is good to go now.

PR #984, opened 2024-04-05
update wb_dims error

olivroy:
Quick fix. Turns out I got here for the first time today! Will finalize next week if necessary. `valid_args_nams` didn't exist. oops

Edit : WIll add test, seems like lacking coverage here https://app.codecov.io/gh/JanMarvin/openxlsx2/pull/984/blob/R/utils.R#L307


JanMarvin:
I know all too well how hard it is to cover all conditions, especially certain `warning()` or `stop()` conditions that you probably never encounter in real life, or at least not in local tests. And although I have my concerns about these 100% coverage projects, many strange things in my code (bugs or at least thinkos) have been discovered using coverage tests, and that's why I'm so eager to always write tests that cover new changes.

JanMarvin (approved):
Looks good to me. Thanks!

PR #986, opened 2024-04-11
Avoid partial matching

olivroy:
caught with options..

Lmk if this is correct or if I missed something!

```r
  options(
    warnPartialMatchArgs = TRUE,
    warnPartialMatchDollar = TRUE,
    warnPartialMatchAttr = TRUE
  )


JanMarvin (approved):
Thanks, looks good! Sorry for the delay, I've been out of action for a bit this week with a cold.

PR #989, opened 2024-04-12
[dims] add check for valid cols argument. fixes #987

JanMarvin:


olivroy:
Looks good. Will be easier to debug this way if a bad input ever gets there!

Issue #988, opened 2024-04-12
Catch incorrect wb_dims


JanMarvin:
Thanks for the report, we will investigate!

@JanMarvin I saw this error in another context, maybe due to this?
https://github.com/JanMarvin/openxlsx2/blob/1194783925e368072b992e042d614a67be6623b5/R/utils.R#L180-L182

_Originally posted by @olivroy in https://github.com/JanMarvin/openxlsx2/issues/985#issuecomment-2049822888_

JanMarvin:
I guess this happens if the input is a dimension without a column attribute? We might have to check if the first character is part of LETTERS (and hope that LETTERS is not localized).

Do you have an example where you ran into this problem, @olivroy ?

olivroy:
not really, it was while dev, never saw it IRL

JanMarvin:
Okay, maybe this is now already covered in `wb_dims()`.

A few strange, but unrelated `wb_dims()` calls ...

```R
> wb_dims(x = mtcars, cols = "25")
[1] "Y2:Y33"
> wb_dims(x = mtcars, cols = "Brrrm")
[1] "BRRRM2:BRRRM33"
> wb_dims(x = data.frame())
[1] "MWLQKWU1:1"
Warning messages:
1: In min(col_int) : no non-missing arguments to min; returning Inf
2: In max(col_int) : no non-missing arguments to max; returning -Inf
```

olivroy:
I can fix those!

JanMarvin:
Thanks, that would be great. I've opened another PR in #989

PR #990, opened 2024-04-12
fix corner cases of wb_dims + standardize error messages

olivroy:
Addresses https://github.com/JanMarvin/openxlsx2/issues/988#issuecomment-2051976050

Doing some touchups locally!

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -280,9 +280,9 @@ check_wb_dims_args <- function(args, select = NULL) {
 
   x_has_colnames <- !is.null(colnames(args$x))
 
-  if (x_has_colnames && !is.null(args$rows) && is.character(args$rows)) {
+  if ( is.character(args$rows) && !is.null(args$rows) && x_has_colnames) {
     # Not checking whether it's a row name, not supported.
-    is_rows_a_colname <- args$row %in% colnames(args$x)
+    is_rows_a_colname <- args$rows %in% colnames(args$x)
 
     if (any(is_rows_a_colname)) {
       stop(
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -842,6 +851,16 @@ wb_dims <- function(..., select = NULL) {
...
+  # patch so that we will want to span at least 1 row and one col
+  if (identical(nrow_to_span, 0L)) {
+    nrow_to_span <- 1L
+  }
+
+  if (identical(ncol_to_span, 0L)) {
+    ncol_to_span <- 1L
+  }
+
   row_span <- frow + seq_len(nrow_to_span) - 1L
```
olivroy, line 864: If `nrow_to_span` was 0, then `seq_len(0)` gave integer(0) frow +integer(0) = integer(0), so the fix just makes sure we span at least one column,

it could also be ncol_to_span <- min(1L, ncol_to_span), would be clearer?

R/utils.R
```diff
@@ -842,6 +851,16 @@ wb_dims <- function(..., select = NULL) {
     }
   }
 
+  # Correct for empty data frame?
```
JanMarvin, line 854: Is this a question? Because I don’t know either 😄

R/utils.R
JanMarvin: It’s fine for me

R/utils.R
```diff
@@ -842,6 +851,16 @@ wb_dims <- function(..., select = NULL) {
...
+  # patch so that we will want to span at least 1 row and one col
```
olivroy, line 855:
```suggestion
  # Ensure we are spanning at least 1 row and 1 col
```

R/utils.R
olivroy: All good. I wrote this before I understood the problem. added a clearer comment.
 so the fix just makes sure we span at least one colum.

JanMarvin: Thanks, I only asked, because sometimes I wonder who wrote a certain comment and why

JanMarvin (approved):
Thanks! Looks like this is good to go

olivroy:
I agree, but at least we have a solid foundation, we know we can trust the results, and won't be able worried to break existing code when we extend it.

PR #991, opened 2024-04-12
Be strict about `cols`, `widths`, `hidden` lengths.

olivroy:
May want to change errors for warnings?

two things here

1. `hidden` is documented as a vector of logical, but class is not enforced, and in tests, integers are provided?
2. I tripped on the length difference of cols and widths

At some point, my script was like that

```r
wb <- wb_workbook()
wb$add_worksheet()
wb$set_col_widths(widths = c(10, 15), cols = c(2, 3))
```

but then I realized that I needed a wider column 4 and I did this:

```r
wb <- wb_workbook()
wb$add_worksheet()
wb$set_col_widths(widths = c(10, 15), cols = c(2, 3, 4))
```

or the opposite, but to realize very much later that it was the wrong thing.

Base R recycling rules are sometimes not what you expect

```r
rep(c(1, 2 ), length.out = 3) # is successful,
#> 1, 2, 1
```

To avoid breaking changes, I could change this to warnings. I used `%%` to only error if one is not a multiple of the other,

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,14 @@
 * Helper to read sensitivity labels from files and apply them to workbooks. 
 [983](https://github.com/JanMarvin/openxlsx2/pull/983)
 
+* It is now possible to pass non consecutive dims like `"A1:B1,C2:D2"` to various style helpers like `wb_add_fill()`. In addition it is now possible to write a data set into a predefined dims region using `enforce = TRUE`. This handles either `","` or `";"` as cell separator. [993](https://github.com/JanMarvin/openxlsx2/pull/993)
+
+```r
+# force a dataset into a specific cell dimension
+wb <- wb_workbook()$add_worksheet()
+wb$add_data(dims = "I2:J2;A1:B2;G5:H6", x = matrix(1:8, 4, 2), enforce = TRUE)
+```
+
 ## Fixes
 
 * Allow writing data frames with zero rows. [987](https://github.com/JanMarvin/openxlsx2/pull/987)
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,12 +17,14 @@ wb_dims(x = mtcars, cols = "non-existent-col")
 # Now errors
 ```
 
-* `wb_set_col_widths()` is more strict about its arguments. If you provide `cols`, `widths`, or `hidden` don't have appropriate length, it will throw an error.
+* `wb_set_col_widths()` is more strict about its arguments. If you provide `cols`, `widths`, or `hidden` don't have appropriate length, it will throw a warning. This may change to an error in the future, so it is recommended to use appropriate values.
 
 
 ***************************************************************************
 
 # openxlsx2 1.5
+
+(This was updated post release.)
 
 ## New features
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,14 @@
 
 * Allow writing data frames with zero rows. [987](https://github.com/JanMarvin/openxlsx2/pull/987)
 
+* `wb_dims()` has been improved and is safer on 0-length inputs. In particular, it will error for a case where a `cols` doesn't exist in `x`
+
+```r
+# Previously created a wrong dims
+wb_dims(x = mtcars, cols = "non-existent-col")
+# Now errors
+```
+
 
 ***************************************************************************
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@
 * Helper to read sensitivity labels from files and apply them to workbooks. 
 [983](https://github.com/JanMarvin/openxlsx2/pull/983)
 
+## Fixes
+
+* Allow writing data frames with zero rows. [987](https://github.com/JanMarvin/openxlsx2/pull/987)
+
+
+***************************************************************************
 
 # openxlsx2 1.5
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,12 +17,12 @@ wb_dims(x = mtcars, cols = "non-existent-col")
 # Now errors
 ```
 
+* `wb_set_col_widths()` is more strict about its arguments. If you provide `cols`, `widths`, or `hidden` don't have appropriate length, it will throw an error.
+
 
 ***************************************************************************
 
 # openxlsx2 1.5
-
-(This was updated post release.)
 
 ## New features
 
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -1278,6 +1278,7 @@ wbWorkbook <- R6::R6Class(
     #' @param na.strings Value used for replacing `NA` values from `x`. Default
     #'   `na_strings()` uses the special `#N/A` value within the workbook.
     #' @param inline_strings write characters as inline strings
+    #' @param enforce enforce that selected dims is filled. For this to work, `dims` must match `x`
     #' @param return The `wbWorkbook` object
     add_data = function(
         sheet            = current_sheet(),
@@ -1295,6 +1296,7 @@ wbWorkbook <- R6::R6Class(
         remove_cell_style = FALSE,
         na.strings        = na_strings(),
         inline_strings    = TRUE,
+        enforce           = FALSE,
         ...
       ) {
 
@@ -1324,7 +1326,8 @@ wbWorkbook <- R6::R6Class(
         apply_cell_style  = apply_cell_style,
         remove_cell_style = remove_cell_style,
         na.strings        = na.strings,
-        inline_strings    = inline_strings
+        inline_strings    = inline_strings,
+        enforce           = enforce
       )
       invisible(self)
     },
@@ -1885,6 +1888,7 @@ wbWorkbook <- R6::R6Class(
     #' @param cm cm
     #' @param apply_cell_style applyCellStyle
     #' @param remove_cell_style if writing into existing cells, should the cell style be removed?
+    #' @param enforce enforce dims
     #' @return The `wbWorkbook` object
     add_formula = function(
         sheet             = current_sheet(),
@@ -1896,6 +1900,7 @@ wbWorkbook <- R6::R6Class(
         cm                = FALSE,
         apply_cell_style  = TRUE,
         remove_cell_style = FALSE,
+        enforce           = FALSE,
         ...
     ) {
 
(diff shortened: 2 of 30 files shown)


NEWS.md
```diff
@@ -17,13 +17,13 @@ wb_dims(x = mtcars, cols = "non-existent-col")
...
 
+* `wb_set_col_widths()` is more strict about its arguments. If you provide `cols`, `widths`, or `hidden` don't have appropriate length, it will throw an error.
+
 
 ***************************************************************************
 
 # openxlsx2 1.5
 
-(This was updated post release.)
-
```
JanMarvin, line 26: Unrelated, buts this should stay, because it significantly differs from the CRAN tar file and the reader should not doubt the tar file.
I don’t care for minor changes like typos or forgotten or fixed links, but here I converted a breaking change to a new feature.

R/class-workbook.R
```diff
@@ -3780,6 +3780,7 @@ wbWorkbook <- R6::R6Class(
       if (length(cols) == 0L) {
         return(self)
       }
+      assert_class(hidden, "logical", arg_nm = "hidden")
```
JanMarvin, line 3783: Will this break existing code? Or is 1L == TRUE?

R/class-workbook.R
```diff
@@ -3791,9 +3792,26 @@ wbWorkbook <- R6::R6Class(
         stop("hidden argument is longer than cols.")
       }
 
+      compatible_length <- length(cols) %% length(widths) == 0
+
+      if (!compatible_length) {
+        # needed because rep(c(1, 2 ), length.out = 3) is successful,
+        # but not clear if this is what the user wanted.
+        stop("`cols` and `widths` don't have compatible lengths.\n",
```
JanMarvin, line 3800: Maybe throw a warning for a while? Before we error out on potentially insignificant errors.

tests/testthat/test-class-workbook.R
```diff
@@ -72,10 +65,20 @@ test_that("wb_set_col_widths", {
...
 
+test_that("set_col_widths errors when inconsistent lengths are supplied", {
+  wb <- wbWorkbook$new()
+  wb$add_worksheet("test")
+
+  expect_error(wb$set_col_widths(cols = c(1, 2, 3), widths = c(2, 3)), "compatible length")
+  expect_error(wb$set_col_widths(cols = "Y", widths = 1:2))
+  expect_error(wb$set_col_widths("test", cols = "Y", hidden = 1:2))
+  expect_error(wb$set_col_widths(cols = c("Y", "Z"), hidden = 1))
+})
```
JanMarvin, line 80: I‘m already away from my computers, but what precisely does hidden > 2 do? And why should a length of 1 not be sufficient? From the last example above I’d guess that columns Y and Z should be hidden. Am I wrong?

JanMarvin (commented):
I agree that seems a bit sketchy. I prefer non breaking changes very much. Preferably they don’t interfere with me as user at all. If I was happy with the code a few months ago, why would I change my mind?

I will think about it over the weekend. Thanks for the PR and the explanations.

olivroy:
No worries, this lacks polishing, but wanted to bake a PR before I didn't care enough.

tests/testthat/test-class-workbook.R
JanMarvin: > hidden (Hidden Columns) 
> Flag indicating if the affected column(s) are hidden on this worksheet.
> The possible values for this attribute are defined by the W3C XML Schema boolean datatype.

From the docs. Either 0/1 or `true`/`false`.

R/class-workbook.R
olivroy: Reverted

tests/testthat/test-class-workbook.R
```diff
@@ -76,6 +69,17 @@ test_that("wb_set_col_widths", {
...
 })
 
+test_that("set_col_widths errors when inconsistent lengths are supplied", {
+  wb <- wbWorkbook$new()
+  wb$add_worksheet("test")
+
+  expect_warning(wb$set_col_widths(cols = c(1, 2, 3), widths = c(2, 3)), "compatible length")
+  expect_error(wb$set_col_widths(cols = "Y", widths = 1:2))
+  expect_error(wb$set_col_widths("test", cols = "Y", hidden = 1:2))
+  expect_warning(wb$set_col_widths(cols = c("X","Y", "Z"), hidden = c(1, 0)))
```
olivroy, line 79:
```suggestion
  expect_warning(wb$set_col_widths(cols = c("X", "Y", "Z"), hidden = c(1, 0)))
```

JanMarvin:
We might have to adjust the changes to test order. It seems that at least one test is now flaking on CI

R/class-workbook.R
```diff
@@ -3780,7 +3780,6 @@ wbWorkbook <- R6::R6Class(
       if (length(cols) == 0L) {
         return(self)
       }
-      assert_class(hidden, "logical", arg_nm = "hidden")
```
JanMarvin: I guess we can relax this, but logical is correct. It can be 0/1 or true/false. Any value greater 0 is interpreted as 1, but provides no additional benefit.

olivroy: openxlsx2:::assert_class(c(0, 1), class = "logical")
#> Error x must be of class logical.

I think we can let it slide for now?

olivroy:
Maybe it is just gh API limitations then..

JanMarvin:
Hi @olivroy , sorry, me opening #1005 was a little late and I didn't want to rush you. I usually strive to announce new releases a bit earlier so that people (well, mostly me) can make time to finish PRs and do further cleanups.

Unfortunately this time I have received a CRAN notification to push a fix for the `isBlankString()` usage and I want to have a little buffer time in case there are other hickups with the submission. (May 1st is labor day in Germany and a national holiday, that's why I picked it.) Just let me know how far this PR is and if it needs additional development time so that we can make time for this.

olivroy:
to me, it feels like it is finished. just adding warnings in the case of incompatible widths / hidden supplied to set_col_widths. 

Leaving the rest unchanged and moving tests in a new entry to catch wb_set_col_widths

It feels less intrusive than throwing error, but signaling that the input may provide unexpected results

tests/testthat/test-class-workbook.R
```diff
@@ -76,6 +69,17 @@ test_that("wb_set_col_widths", {
 
 })
 
+test_that("set_col_widths errors when inconsistent lengths are supplied", {
+  wb <- wbWorkbook$new()
+  wb$add_worksheet("test")
+
+  expect_warning(wb$set_col_widths(cols = c(1, 2, 3), widths = c(2, 3)), "compatible length")
+  expect_error(wb$set_col_widths(cols = "Y", widths = 1:2))
+  expect_error(wb$set_col_widths("test", cols = "Y", hidden = 1:2))
```
JanMarvin, line 78:
```suggestion
expect_error(wb$set_col_widths("test", cols = "Y", hidden = 1:2), "hidden argument is longer than cols.")
```

tests/testthat/test-class-workbook.R
```diff
@@ -76,6 +69,17 @@ test_that("wb_set_col_widths", {
 
 })
 
+test_that("set_col_widths errors when inconsistent lengths are supplied", {
+  wb <- wbWorkbook$new()
+  wb$add_worksheet("test")
+
+  expect_warning(wb$set_col_widths(cols = c(1, 2, 3), widths = c(2, 3)), "compatible length")
+  expect_error(wb$set_col_widths(cols = "Y", widths = 1:2))
```
JanMarvin, line 77:
```suggestion
expect_error(wb$set_col_widths(cols = "Y", widths = 1:2), "More widths than columns supplied.")
```

tests/testthat/test-class-workbook.R
```diff
@@ -76,6 +69,17 @@ test_that("wb_set_col_widths", {
 
 })
 
+test_that("set_col_widths errors when inconsistent lengths are supplied", {
```
JanMarvin, line 72: Maybe `"set_col_widths informs when inconsistent lengths are supplied"`?
But that's just nitpicking

JanMarvin (approved):
Thanks, I just now realize that you have simply moved the `expect_error()` tests, but I actually feel that testing against some condition is worth it (we once had a `expect_warning()` test that succeeded, but the warning was produced for reasons unrelated to `openxlsx2` and not what was supposed to be covered 😀 

Otherwise good to go!

tests/testthat/test-class-workbook.R
```diff
@@ -76,6 +69,16 @@ test_that("wb_set_col_widths", {
...
 })
 
+test_that("set_col_widths informs when inconsistent lengths are supplied", {
+  wb <- wbWorkbook$new()
+  wb$add_worksheet("test")
+
+  expect_warning(wb$set_col_widths(cols = c(1, 2, 3), widths = c(2, 3)), "compatible length")
+  expect_error(wb$set_col_widths(cols = "Y", widths = 1:2), "More widths than column")
+  expect_error(wb$set_col_widths("test", cols = "Y", hidden = 1:2), "hidden argument is longer")
+  expect_warning(wb$set_col_widths(cols = c("X","Y", "Z"), hidden = c(1, 0)), "compatible length")
```
olivroy, line 79:
```suggestion
  expect_warning(wb$set_col_widths(cols = c("X", "Y", "Z"), hidden = c(1, 0)), "compatible length")
```

olivroy:
Thanks! So, I guess you can merge and then test revdeps?

R/class-workbook.R
```diff
@@ -16,7 +16,7 @@
 #' ``` r
 #' wb <- wb_workbook(creator = "My name here") %>%
 #'   wb_add_worksheet(sheet = "Expenditure", grid_lines = FALSE) %>%
-#'   wb_add_data(x = USPersonalExpenditure, row_names = TRUE)
+#'   wb_add_data(x = USPermsonalExpenditure, row_names = TRUE)
```
JanMarvin, line 19: Was this intended?

JanMarvin:
Please fix the lintr warning and the probably unintended change in the example code. Feel free to squash merge afterwards. Thanks!

R/class-workbook.R
olivroy: I am so silly..
```suggestion
#'   wb_add_data(x = USPersonalExpenditure, row_names = TRUE)
```

olivroy:
Fixed both the lintr warning and the typo.. Sometimes RStudio types things for me :(

PR #992, opened 2024-04-12
Partial match

olivroy:
Just so I don't send them in small pieces, I decided to do a scrubbing.

+ changed running order of tests, since read_sources and class-workbook-wrappers are the slowest tests

### DIFF
--- a/R/helper-functions.R
+++ b/R/helper-functions.R
@@ -824,7 +824,7 @@ get_colstyle <- function(wb, sheet = current_sheet()) {
   st_ids <- NULL
   if (length(wb$worksheets[[sheet]]$cols_attr)) {
     cols <- wb$worksheets[[sheet]]$unfold_cols()
-    st_ids <- cols$s[cols$s != ""]
+    st_ids <- cols$style[cols$style != ""]
     xf_ids <- match(st_ids, wb$styles_mgr$xf$id)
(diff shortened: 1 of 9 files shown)


R/helper-functions.R
```diff
@@ -824,7 +824,7 @@ get_colstyle <- function(wb, sheet = current_sheet()) {
   st_ids <- NULL
   if (length(wb$worksheets[[sheet]]$cols_attr)) {
     cols <- wb$worksheets[[sheet]]$unfold_cols()
-    st_ids <- cols$s[cols$s != ""]
+    st_ids <- cols$style[cols$style != ""]
```
JanMarvin, line 827: Not sure about this. This is in an unfolded column.  In rows they use s for styles, but it could be corrected. I’ll look it up. The rest is fine, thanks!

olivroy, line 827:
```suggestion
    st_ids <- cols$s[cols$s != ""]
```

JanMarvin: Alright, it’s [style](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.columns?view=openxml-3.0.1)

JanMarvin: You were correct in the first place

olivroy: Sorry I went too fast. Will merge when CI green

Issue #796, opened 2023-09-17
Improve dims


JanMarvin:
Due to some merged cell styles sometimes something like `A1;B2:C2;A3:C5` is needed to import to a data frame. Sadly this is currently not possible.

Our selection of multiple cells always creates a square from the most left to the most right cell:

``` r
library(openxlsx2)

x <- matrix(1:9, 3, 3)

wb <- wb_workbook()$add_worksheet()$add_data(x = x, col_names = FALSE)

wb_to_df(wb, dims = "A1;B2;C3", col_names = FALSE)
#>   A B C
#> 1 1 4 7
#> 2 2 5 8
#> 3 3 6 9
```

It would be beneficial if this would return `1, 5, 8` instead, or a data frame where all other cells are `NA`. But if I remember correctly this requires at least modifications of `dims_to_dataframe()` and the question remains if this `"A1;B2;C3"` should return a vector or a data frame of a single row or ... something entirely different.

JanMarvin:
Yes I do. But to make it work consistently, some development and lots of testing has to be done. After all there is a reason why it doesn't work right now

JanMarvin:
I was thinking about a data validation book chapter and was wondering if it makes sense to add sheet names to `wb_dims()`. For this `wb_dims()` should become a workbook wrapper, since it needs to know the sheet names.

Draft:
```R
wb$dims(sheet = "MySheet", x = mtcars)
#> MySheet!A1:K33

wb$dims(sheet = "My Sheet", x = mtcars)
#> 'My Sheet'!A1:K33
```

If we grep for `!` in `dims` we could string split `sheet` and `dims` from there. In the current data validation list example we use a spreadsheet reference: `"'Sheet 2'!$A$1:$A$10"`. This could be provided by something like: `wb$dims(sheet = 2, cols = 1, rows = seq_len(10))`. But I'm not entirely sure where else we would benefit from this instead of potentially being closer to openxml formulas.

The following functions would be identical:
```R
wb$to_df(dims = wb$dims(sheet = "Sheet2", x = matrix(1, 2, 2), col_names = FALSE)
wb$to_df(dims = "Sheet2!A1:B2")
wb_to_df(wb, sheet = "Sheet2", dims = wb_dims(wb = wb, x = matrix(1, 2, 2), col_names = FALSE)
wb_to_df(wb, sheet = "Sheet2", dims = "A1:B2")
```

The data validation example would become something like this:
```R
# data validation example
[...]

PR #993, opened 2024-04-13
[WIP] More flexible dims

JanMarvin:
Still a little hackish solution to the flexibility problem mentioned in #796 , but it shows what is possible. For whatever reason the enforce option does not work with `wb_add_fill()` probably because empty cells are still initialized even though they are not filled.

Also there is still a bug, for now I simply assumed that all data columns are beneath each other to fill their positions. In the example below the column header is right to the data and hence the row has to be filled wider, which is not yet implemented. Hence the first data part is filled incorrectly.

``` r
library(openxlsx2)

# write a workbook with a few colored cells
wb <- wb_workbook()$
  add_worksheet()$
  add_fill(dims = "A1:B2;C3:D4;E1:F2;G5:H6;I2:K2", color = wb_color("yellow"))

# force a dataset into a specific cell dimension
options("openxlsx2.enforce_dims" = TRUE)
wb$add_data(dims = "I2:J2;A1:B2;G5:H6", x = matrix(1:8, 4, 2))
options("openxlsx2.enforce_dims" = FALSE)

if (interactive()) wb$open()
```

Screenshot 2024-04-13 at 13 26 30

### DIFF
--- a/R/wb_functions.R
+++ b/R/wb_functions.R
@@ -45,10 +45,10 @@ dims_to_dataframe <- function(dims, fill = FALSE) {
   }
 
   dims_to_df(
-    rows = rows_out,
-    cols = cols_out,
+    rows   = rows_out,
+    cols   = cols_out,
     filled = filled,
-    fill = fill
+    fill   = fill
   )
 }
 
--- a/R/wb_functions.R
+++ b/R/wb_functions.R
@@ -11,6 +11,9 @@ dims_to_dataframe <- function(dims, fill = FALSE) {
   if (grepl(";", dims)) {
     dims <- unlist(strsplit(dims, ";"))
   }
+  if (grepl(",", dims)) {
+    dims <- unlist(strsplit(dims, ","))
+  }
 
   rows_out <- NULL
   cols_out <- NULL
--- a/R/wb_functions.R
+++ b/R/wb_functions.R
@@ -44,22 +44,12 @@ dims_to_dataframe <- function(dims, fill = FALSE) {
     }
   }
 
-  if (is.null(filled)) {
-    # create data frame from rows/
-    dims_to_df(
-      rows = rows_out,
-      cols = cols_out,
-      fill = fill
-    )
-  } else {
-    # create data frame from rows/
-    dims_to_df2(
-      rows   = rows_out,
-      cols   = cols_out,
-      filled = filled,
-      fill   = fill
-    )
-  }
+  dims_to_df(
+    rows = rows_out,
+    cols = cols_out,
+    filled = filled,
+    fill = fill
+  )
 }
 
 #' Create dimensions from dataframe
--- a/R/wb_functions.R
+++ b/R/wb_functions.R
@@ -76,28 +76,26 @@ dims_to_dataframe <- function(dims, fill = FALSE) {
 dataframe_to_dims <- function(df, dim_break = TRUE) {
 
   if (dim_break) {
-    # get continuous sequences of columns and rows in df
-    v <- as.integer(rownames(df))
-    rows <- split(v, cumsum(diff(c(-Inf, v)) != 1))
 
-    v <- col2int(colnames(df))
-    cols <- split(colnames(df), cumsum(diff(c(-Inf, v)) != 1))
+    dims <- dims_to_dataframe(dataframe_to_dims(df, dim_break = FALSE), fill = TRUE)
 
-    # combine columns and rows to construct dims
-    out <- NULL
-    for (col in seq_along(cols)) {
-      for (row in seq_along(rows)) {
-        tmp <- paste0(
-          cols[[col]][[1]], rows[[row]][[1]],
-          ":",
-          rev(cols[[col]])[[1]],  rev(rows[[row]])[[1]]
-        )
-        out <- c(out, tmp)
-      }
-    }
+    mm <- as.matrix(df)
+    mm[mm != "" | is.na(mm)] <- 1
+    mm[mm == ""] <- 0
 
-    return(paste0(out, collapse = ";"))
+    matrix <- matrix(as.numeric(mm), nrow(mm), ncol(mm))
+    dimnames(matrix) <- list(rownames(mm), colnames(mm))
+
+    # remove columns and rows not in df
+    dims <- dims[, colnames(dims) %in% colnames(matrix)]
+    dims <- dims[rownames(dims) %in% rownames(matrix), ]
+
+    out <- dims[matrix == 1]
+
+    return(paste0(out, collapse = ","))
+
   } else {
+
     rows <- as.integer(rownames(df))
     cols <- colnames(df)
 
@@ -108,6 +106,7 @@ dataframe_to_dims <- function(df, dim_break = TRUE) {
     )
 
     return(tmp)
+
   }
 }
 
--- a/R/wb_functions.R
+++ b/R/wb_functions.R
@@ -66,34 +66,48 @@ dims_to_dataframe <- function(dims, fill = FALSE) {
 #'
 #' Use [wb_dims()]
 #' @param df dataframe with spreadsheet columns and rows
+#' @param dim_break split the dims?
 #' @examples
 #'  df <- dims_to_dataframe("A1:D5;F1:F6;D8", fill = TRUE)
 #'  dataframe_to_dims(df)
 #' @keywords internal
 #' @export
-dataframe_to_dims <- function(df) {
+dataframe_to_dims <- function(df, dim_break = TRUE) {
 
-  # get continuous sequences of columns and rows in df
-  v <- as.integer(rownames(df))
-  rows <- split(v, cumsum(diff(c(-Inf, v)) != 1))
+  if (dim_break) {
+    # get continuous sequences of columns and rows in df
+    v <- as.integer(rownames(df))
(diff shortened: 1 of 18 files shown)


tests/testthat/test-wb_functions.R
```diff
@@ -188,11 +188,12 @@ test_that("dims_to_dataframe", {
...
-  df <- dims_to_dataframe("A1:D5;F1:F6;D8", fill = TRUE)
-  dims <- dataframe_to_dims(df)
-  df2 <- dims_to_dataframe(dims, fill = TRUE)
-  expect_equal(df, df2)
+  ## I should probably fix this failing test
+  # # dims_to_dataframe will always create a square
+  # df <- dims_to_dataframe("A1:D5;F1:F6;D8", fill = TRUE)
+  # dims <- dataframe_to_dims(df)
+  # df2 <- dims_to_dataframe(dims, fill = TRUE)
+  # expect_equal(df, df2)
```
JanMarvin, line 196: This test is not so important. It wasn't working entirely correct previously, it simply created two data frames of similar size, but did not care about not selected cells

src/helper_functions.cpp
```diff
@@ -237,6 +237,59 @@ SEXP copy(SEXP x) {
   return Rf_duplicate(x);
 }
 
+// FIXME C++17 is quite a jump for such a tiny function ...
```
JanMarvin, line 240: there should be a c++11 solution, but this was the first I found

JanMarvin:
``` r
library(openxlsx2)

# write a workbook with a few colored cells
wb <- wb_workbook()$
  add_worksheet()$
  add_fill(dims = "A1:B2;C3:D4;E1:F2;G5:H6;I2:K2", color = wb_color("yellow"))

# force a dataset into a specific cell dimension
options("openxlsx2.enforce_dims" = TRUE)
wb$add_data(dims = "I2:J2;A1:B2;G5:H6", x = matrix(1:8, 4, 2))
options("openxlsx2.enforce_dims" = FALSE)

wb$add_font(dims = "B2;G6;J2", bold = TRUE)

if (interactive()) wb$open()
```

Screenshot 2024-04-13 at 19 09 48


Not sure if this has any impact on our usual behavior. Enforce only works with data, therefore it might be a good idea to add it as option.

R/wb_functions.R
```diff
@@ -40,46 +44,70 @@ dims_to_dataframe <- function(dims, fill = FALSE) {
...
+      fill = fill
+    )
+  } else {
+    # create data frame from rows/
+    dims_to_df2(
+      rows   = rows_out,
+      cols   = cols_out,
+      filled = filled,
+      fill   = fill
+    )
```
JanMarvin, line 61: The sort order is not necessarily correct. Should be fixed. Otherwise selecting first column name and first row name might provide incorrect results:
```R
> dims_to_dataframe("I2:J2;A1:B2;G5:H6")
     I    J    A    B    G    H
2 <NA> <NA> <NA> <NA>          
1           <NA> <NA>          
5                     <NA> <NA>
6                     <NA> <NA>
```

R/wb_functions.R
```diff
@@ -40,46 +44,70 @@ dims_to_dataframe <- function(dims, fill = FALSE) {
...
+    out <- NULL
+    for (col in seq_along(cols)) {
+      for (row in seq_along(rows)) {
+        tmp <- paste0(
+          cols[[col]][[1]], rows[[row]][[1]],
+          ":",
+          rev(cols[[col]])[[1]],  rev(rows[[row]])[[1]]
+        )
+        out <- c(out, tmp)
+      }
```
JanMarvin, line 95: For `dim_break` this code creates incorrect results.

R/write.R
```diff
@@ -299,7 +302,10 @@ write_data2 <- function(
     data <- as.data.frame(t(data))
   }
 
-  dims <- fits_in_dims(x = data, dims = dims, startCol = startCol, startRow = startRow)
+  # hackish solution
+  if (!getOption("openxlsx2.enforce_dims", default = FALSE)) {
+    dims <- fits_in_dims(x = data, dims = dims, startCol = startCol, startRow = startRow)
+  }
```
JanMarvin, line 308: This option is a little all over the place not a huge fan

R/write.R
```diff
@@ -932,7 +957,7 @@ write_data_table <- function(
     na.strings      = na.strings,
     data_table      = data_table,
     inline_strings  = inline_strings,
-    dims            = dims
+    dims            = odims
```
JanMarvin, line 960: at this stage `dims` is the result of `dims_to_rowcol()` iirc. I overwrote whatever else was going in the code above. Might or might not be needed anymore. Dunno. The write-code is a bottomless pit and you better don't stare to long in it ...

src/helper_functions.cpp
```diff
@@ -237,6 +237,76 @@ SEXP copy(SEXP x) {
...
+    std::string result;
+    for (char c : str) {
+        if (!std::isdigit(c)) {
+            result += c;
+        }
+    }
+    return result;
+}
+
+std::string rm_alf(const std::string& str) {
```
JanMarvin, line 250: Alf

src/helper_functions.cpp
```diff
@@ -237,6 +237,76 @@ SEXP copy(SEXP x) {
   return Rf_duplicate(x);
 }
 
+std::string rm_dig(const std::string& str) {
```
JanMarvin, line 240: I lowkey believe that there are already functions to do just this ...

JanMarvin:
@olivroy This PR is closing in, just minor things from my side. It would be nice if the disabled test would work, but at the moment I’m not in the mood to figure this out. Guess we can live without it, at least for the moment. Returning different combinations of (squared) ranges and cells is a little tricky.

But finally we can and probably should extend `wb_dims()` with this PR. We should be able to select multiple columns from a data frame, but we might be able to extend it even further. 

1) Maybe provide column and row values, but return only specific cells
```r
wb_dims(cols = 1:2, rows = c(1,3), cells = TRUE)
#> “A1,B3”
```
or 

2) highlight a dataframe based on a condition
```r
wb_dims(x = mtcars[mtcars > 4], cells = TRUE)
```

Making `wb_dims()` more flexible has been requested on [SO](https://stackoverflow.com/a/78349187/12340029)

What might be an issue is speed. I’ve tested this PR only with small dims, only a few cells not hundreds or thousands, but we’ll see…

R/write.R
```diff
@@ -216,9 +220,16 @@ write_data2 <- function(
     na.strings = na_strings(),
     data_table = FALSE,
     inline_strings = TRUE,
-    dims = NULL
+    dims = NULL,
```
olivroy, line 223: Maybe adding dims to `write_xlsx()` could be its own PR if there are too manu moving parts. Would need to check `rows`, `cols`, `start_row` etc. ?

R/write.R
```diff
@@ -1059,6 +1097,7 @@ write_data <- function(
     remove_cell_style = FALSE,
     na.strings        = na_strings(),
     inline_strings    = TRUE,
+    enforce           = FALSE,
```
olivroy, line 1100: Adding another argument everywhere seems a bit verbose..

I don't know if this is good, but possibly via options is better? or 

adding it to user facing function save in temp env var, then apply it where needed?

Probably not a good suggestion

olivroy (commented):
Seems interesting!

So if I understand correctly, `dims = "A1;B2:C3"` didn't work before this PR?

R/write.R
JanMarvin: If I understand you correctly: I made it an argument, because with the option you have to actively set it if needed and it didn’t work with fill either way.

JanMarvin:
> Seems interesting!
> 
> So if I understand correctly, `dims = "A1;B2:C3"` didn't work before this PR?

Yes previously this didn’t work and now it will work only if you have a scalar and a matrix with 4 values. There are obviously cornerstones that I haven’t figured out yet, but this PR is a huge step in the right direction.

JanMarvin:
``` r
library(openxlsx2)
  add_data(dims = "A1;B2:C3", x = 1:5, enforce = TRUE)
```

This works too, but I see the strength of different dim ranges more in evenly stretched columns:

``` r
library(openxlsx2)
wb <- wb_workbook()$
  add_worksheet()$
  add_data(dims = "A1", x = mtcars)$
  add_fill(dims = wb_dims(x = mtcars, cols = c("cyl", "wt")), color = wb_color("green"))
```

Where we previously could only color a single column, we are now able to color both. If `wb_dims(x = mtcars, cols = c("cyl", "wt"))` returns something like `"B2:B33,F2:F33"`.

R/write.R
JanMarvin: Interesting, I wasn't aware that we do not support `dims` in `write_xlsx()`, but maybe that's not so important. Presumably people that use `write_xlsx()` only want a data frame or vector written as xlsx file and do not care that much about styling.

JanMarvin:
This works:

```R
  add_data(dims = "A1:B5", x = matrix(1:10, ncol = 2),
           enforce = TRUE, col_names = FALSE)

  add_data(dims = "A1:B1;A3:B6", x = matrix(1:10, ncol = 2),
           enforce = TRUE, col_names = FALSE)
```

This does not work:
```R
  add_data(dims = "A1:A5,B1:B5", x = matrix(1:10, ncol = 2),
           enforce = TRUE, col_names = FALSE)
```

PR #994, opened 2024-04-15
[gha] add recheck.yml

JanMarvin:


JanMarvin:
Thank you very much @jeroen! Just two weeks ago I was wishing something like this could be solved easier when I was wrangling with my previous revdepcheck workflow, and now you just did it :)

JanMarvin:
Yes, it succeeded. Tried both: strong and most.

PR #1015, opened 2024-05-09
[dims] previously non-consecutive dims could get lost

### LINKED ISSUE #1014: Non-Consecutive dim gets lost
``` r
library(openxlsx2)

# dims <- "B7:B9,C6:C10,D5:D11,E5:E12,F6:F13,G7:G14,H6:H13,I5:I12,J5:J11,K6:K10,L7:L9"
dims <- "B7:B9,C6:C10,D5:D11,E5:E12,G7:G14,H6:H13,I5:I12,J5:J11,K6:K10,L7:L9"

wb <- wb_workbook()$
  add_worksheet(grid_lines = FALSE)$
  set_col_widths(cols = "A:M", widths = 2)$
  add_fill(dims = dims, color = wb_color("red"))


if (interactive()) wb$open()
```

I was toying around with coloring non consecutive cell ranges to create a heart shape using this as reference: 
Reference Heart shape

In my initial approach I missed a column, but to my surprise, column L is missing too (picture on the left). Even though it is in the code. If I use the full range, the image looks fine (picture on the right).
Screenshot 2024-05-09 at 13 37 29 Screenshot 2024-05-09 at 13 38 03

JanMarvin:
if rows and columns were entirely omitted. This could create a cc object that was to short. fixes #1014

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,10 @@
 
 * Silence a warning triggered by a folder called `"[trash]"`. [1012](https://github.com/JanMarvin/openxlsx2/pull/1012)
 
+## Fixes
+
+* Fixed an issue with non consecutive dims, where columns or rows were silently dropped. [1015](https://github.com/JanMarvin/openxlsx2/pull/1015)
+
 
 ***************************************************************************
 
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -9151,7 +9151,7 @@ wbWorkbook <- R6::R6Class(
         if (length(need_cells) == 1 && grepl(":|;", need_cells))
           need_cells <- dims_to_dataframe(dims, fill = TRUE)
 
-        exp_cells <- unname(unlist(need_cells))
+        exp_cells <- unname(unlist(need_cells[need_cells != ""]))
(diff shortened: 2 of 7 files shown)


JanMarvin:
It's not the proudest thing I've ever programmed, but it should be better than the previous approach.

The problem is that the length of the dims of the data to be selected and the entire data range are unequal. 
Imagine a case where dims skips a row, as in `“A1:B2,A4:B6”`. In some places, we simply loop over the number of columns and rows. If a row is missing, our loops are not aware of it. There is nothing to do for row 3, but they iterate over it and are at n+1 and therefore later stop at n-1. The correction therefore enlarges the selected data frame so that it covers the entire data range. This way, when looping over the data, both sizes are equal.

Unfortunately, in these cases we might have to create a few empty columns and rows.

The whole code is unfortunately a little ... confusing. I'm not going to call it spaghetti code just yet, but it's getting close due to the various stages of development we're still supporting. One day `write_data()` and `write_data2()` should be combined and the whole thing tidied up. We can also iterate over the necessary rows and columns. But that will probably only happen in the distant future.

PR #1018, opened 2024-05-14
[vignette] begin redesign of basic manual

JanMarvin:


JanMarvin:
It's work in progress, but probably a bit better than the previous long list of read arguments.

PR #1019, opened 2024-05-18
[dims] extend wb_dims() to accept multiple columns

JanMarvin:
Fix #843

This PR extends `wb_dims()` to accept multiple columns:

``` r
library(openxlsx2)
wb <- wb_workbook() %>% 
  wb_add_worksheet() %>% 
  wb_add_data(x = mtcars) %>% 
  wb_add_fill(dims = wb_dims(x = mtcars, cols = c("cyl", "gear"), select = "col_names"), color = wb_color("blue")) %>% 
  wb_add_fill(dims = wb_dims(x = mtcars, cols = c("cyl", "gear")), color = wb_color("orange"))

if (interactive())  wb %>% wb_open()
```

Screenshot 2024-05-18 at 13 54 04

``` r
# just to show what is going on
wb_dims(x = mtcars, cols = c("cyl"))
#> [1] "B2:B33"

wb_dims(x = mtcars, cols = c("cyl", "gear"))
#> [1] "B2:B33,J2:J33"

wb_dims(x = mtcars, select = "col_names")
#> [1] "A1:K1"

wb_dims(x = mtcars, cols = c("cyl", "gear"), select = "col_names")
#> [1] "B1,J1"
```

PR #1022, opened 2024-05-22
Clearer message for bad select in wb_dims()

olivroy:
Addresses https://github.com/JanMarvin/openxlsx2/pull/1019#issuecomment-2118822103

Nudge towards using `row_names = TRUE` inside `wb_dims()`. Let me know if you like that.

The goal is also to avoid confusion, because mtcars has row names, but if you don't provide `row_names = TRUE` in a couple places in openxlsx2, they will be stripped.

```r
# before
wb_dims(x = mtcars, select = "row_names")
#>     You provided a bad value to `select` in `wb_dims()`.
 #>      Please review. see `?wb_dims`.
# this PR
wb_dims(x = mtcars, select = "row_names")
#>     `select` can't be "row_names" if `x` doesn't have row names.
 #>      Use `row_names = TRUE` inside `wb_dims()` to make sure row names are preserved.
```


JanMarvin:
Thanks, I don’t have a lot of time today, but generally the idea is fine. It’s now clear that there is something else required, whereas previously the input looked fine and there was no obvious input error.

I assume you’re not a huge fan, of this but `wb_dims()` could pick up `col_names` from the parent environment? But as usually I haven’t thought about strange consequences of this.

JanMarvin (approved):
Thanks! Looks good 👍

olivroy:
> I assume you’re not a huge fan, of this but wb_dims() could pick up col_names from the parent environment? But as usually I haven’t thought about strange consequences of this.

That would scare me indeed! would have to find previous art (in openxlsx2 or elsewhere) to be confident with implementation.

But that would require the concept of each sheet default `dims`

dims = current_sheet_dims() a waiver ?

Issue #1032, opened 2024-05-30
wb_save outputs a file which Excel perceives as corrupted


itsmevictor:
My issue is that regardless of the content of the workbook and regardless of the modifications I perform on it, the file outputted by `wb_save` is perceived as corrupted by Excel (but not Libre Office). 

I ended up testing with an empty .xlsx file containing only the value 1 in the cell A1 and ran the following (that is without performing any modifications on the workbook): 
```
template_file <- openxlsx2::wb_load(file = paste0(wd, "/data/template_liste.xlsx"))

openxlsx2::wb_save(wb = template_file, file = paste0(wd, "/data/test_template_liste.xlsx"), overwrite = TRUE)
```
but for some reason, when trying to open the file in Excel, I get the following error message _"We found a problem with some content in 'test_template_liste.xlsx'. Do you want us to try to recover as much as we can? If you trust the source of this workbook, click Yes."._. If I click Yes, the content of the repair is the following:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
> <recoveryLog xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"><logFileName>Repair Result to test_template_liste6.xml</logFileName><summary>Errors were detected in file '/data/test_template_liste.xlsx'</summary><additionalInfo><info>Excel completed file level validation and repair. Some parts of this workbook may have been repaired or discarded.</info></additionalInfo><repairedParts summary="Following is a list of repairs:"><repairedPart>Repaired Part: /xl/workbook.xml part with XML error.  (Workbook) Xml parsing error Line 1, column 1679.</repairedPart></repairedParts></recoveryLog>
```

This repair message appears to be the same regardless of the content of the workbook. Noteworthily, I do not have the same issue with LibreOffice *and* I do not have the same issue with Python's openpyxl (including for Excel). 

I'm using Excel 16.85 on Mac OS and I'm using R 4.4.0 in Visual Studio Code 1.89.1. An example of a corrupted Excel file is here: 
[test_template_liste.xlsx](https://github.com/JanMarvin/openxlsx2/files/15501830/test_template_liste.xlsx)

I hope the issue is clear enough. Many thanks in advance for your help!

JanMarvin:
Hi @itsmevictor , Thanks for the report, I’ll look into it.

Two things
* Can you please provide the file prior to our interaction with it? The file you want to load and save.
* Does it work if you save a file not loaded? Like `write_xlsx()`

[...]

PR #1033, opened 2024-05-31
[blob] read blob found in xl folder

JanMarvin:
This handles the blob found in #1032

### DIFF
--- a/R/wb_load.R
+++ b/R/wb_load.R
@@ -207,16 +207,6 @@ wb_load <- function(
   ## VBA Macro
   vbaProject        <- grep_xml("vbaProject\\.bin$")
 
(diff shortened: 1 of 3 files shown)


R/wb_load.R
```diff
@@ -1674,6 +1686,25 @@ wb_load <- function(
     wb$workbook$xti <- NULL
   }
 
+  do_not_remove <- c(
+    if (length(wb$blob)) paste0(basename2(wb$blob)) else NULL,
```
JanMarvin, line 1690: Should be: `paste0(basename2(wb$blob), collapse = "|")`, because ws$blob could contain multiple blobs

PR #1035, opened 2024-06-01
fix warnings from implicit conversions

JanMarvin:
`-Wconversion -Wno-sign-conversion`

### DIFF
--- a/src/xlsb_funs.h
+++ b/src/xlsb_funs.h
@@ -4,7 +4,9 @@
 #include <string>
 #include <fstream>
 #include <cstdint>
-#include <typeinfo>
+
+// for swap_endian
+#include <type_traits>
 
 // detect if we need to swap. assuming that there is no big endian xlsb format,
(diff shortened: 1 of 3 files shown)


src/xlsb_funs.h
```diff
@@ -188,13 +207,13 @@ std::string read_xlwidestring(std::string &mystring, std::istream& sas) {
...
   std::u16string str;
-  str.resize((double)size * 2);
+  str.resize(size * 2);
 
   if (!sas.read((char*)&str[0], str.size()))
     Rcpp::stop("char: a binary read error occurred");
 
   std::string outstr = to_utf8(str);
-  if ((double)str.size()/2 != size) Rcpp::warning("String size unexpected");
+  if (str.size()/2 != size) Rcpp::warning("String size unexpected");
```
JanMarvin, line 216: A little bit unsure about this. Iirc I ran into an overflow issue here, but I'm unable to remember if this was caused by `size_t` or something else.

Issue #1060, opened 2024-06-27
Roadmap for release 1.8


JanMarvin:
This will be release 1.8 and should be the 20th released to CRAN.

I'm planning on a release maybe at the end of next week.

## open pull requests to merge
* N/A

## open pull requests to modify
* N/A

## issues to work on pre release
* N/A

olivroy:
Would now be a good time to make the `write_formula()`, `write_comment()`, `write_data()` etc. internal?

JanMarvin:
Indeed, I have thought about this too. Maybe move `write_data()` to `do_write_data()` and create something like:

```R
write_data <- function(...) {
  .Deprecated("wb_add_data", package = "openxlsx2", old = "write_data")
  do_write_data(...)
}
```

olivroy:
I opened an issue in https://github.com/Public-Health-Scotland/SBSS/issues/55, but without response.

JanMarvin:
I'll try to create a PR over the weekend. My idea would be to throw a deprecation warning for some time and remove the function afterwards. (Not a big fan of removing publicly available APIs on short notice, even though it's quite common in the R world).

PR #1064, opened 2024-06-27
Deprecate old functions `write_*()` and `remove_*()`

olivroy:
`remove_comment()` `write_comment()`, `write_formula()`, `write_datatable()`, `convertToExcelDate()` -> `convert_to_excel_date()`

#1060


tests/testthat/test-wb_functions.R
```diff
@@ -34,7 +34,7 @@ test_that("wb_to_df", {
   expect_equal(convert_to_excel_date(df = exp["Var5"], date1904 = FALSE),
                got["Var5"])
 
-  expect_warning(convertToExcelDate(df = exp["Var5"], date1904 = FALSE), "deprecated")
+  expect_error(expect_warning(convertToExcelDate(df = exp["Var5"], date1904 = FALSE), "deprecated"),"defunct")
```
olivroy, line 37:
```suggestion
  expect_error(expect_warning(convertToExcelDate(df = exp["Var5"], date1904 = FALSE), "deprecated"), "defunct")
```

JanMarvin:
Thanks! Looking good already, but it’s 1pm here and I’m going to sleep now.

Nitpicking: Please move the old functions below the do functions. Just so the code can be read from top.

olivroy:
moved all old functions to own file. wdyt?

R/deprecate.R
```diff
@@ -0,0 +1,227 @@
...
+    banded_rows       = TRUE,
+    banded_cols       = FALSE,
+    apply_cell_style  = TRUE,
+    remove_cell_style = FALSE,
+    na.strings        = na_strings(),
+    inline_strings    = TRUE,
+    total_row         = FALSE,
+    ...
+) {
+  .Deprecated("wb_add_data_table()", package = "openxlsx2", old = "write_datatable")
```
JanMarvin, line 160: "write_datatable()"

R/deprecate.R
```diff
@@ -0,0 +1,227 @@
...
+    start_col         = 1,
+    start_row         = 1,
+    array             = FALSE,
+    cm                = FALSE,
+    apply_cell_style  = TRUE,
+    remove_cell_style = FALSE,
+    enforce           = FALSE,
+    ...
+) {
+  .Deprecated("wb_add_formula()", package = "openxlsx2", old = "write_formula")
```
JanMarvin, line 211: "write_formula()"

R/deprecate.R
```diff
@@ -0,0 +1,227 @@
...
+
+#' Convert to Excel data
+#'
+#' Use [convert_to_excel_date()].
+#' @usage NULL
+#' @inheritParams convert_to_excel_date
+#' @keywords internal
+#' @export
+convertToExcelDate <- function(df, date1904 = FALSE) {
+  .Deprecated(old = "convertToExcelDate", new = "convert_to_excel_date", package = "openxlsx2")
```
JanMarvin, line 73: guess we can remove this line? The code will stop on the next either way. Should we still export this? I would simply no longer export this and drop the function after a while.

tests/testthat/test-wb_functions.R
```diff
@@ -34,7 +34,7 @@ test_that("wb_to_df", {
   expect_equal(convert_to_excel_date(df = exp["Var5"], date1904 = FALSE),
                got["Var5"])
 
-  expect_warning(convertToExcelDate(df = exp["Var5"], date1904 = FALSE), "deprecated")
+  expect_error(expect_warning(convertToExcelDate(df = exp["Var5"], date1904 = FALSE), "deprecated"), "defunct")
```
JanMarvin, line 37: Guess we can drop the deprecated and simply throw defunct (see above)

JanMarvin (approved):
Thanks, left some minor remarks.

R/deprecate.R
olivroy: If you stop exporting. People using it will get function not found, which is basically equivalent to dropping it fully (which I don't mind doing) I couldn't find any usage on GitHub.

olivroy, line 73:
```suggestion
```

R/deprecate.R
olivroy, line 160:
```suggestion
  .Deprecated("wb_add_data_table()", package = "openxlsx2", old = "write_datatable()")
```

R/deprecate.R
olivroy, line 211:
```suggestion
  .Deprecated("wb_add_formula()", package = "openxlsx2", old = "write_formula()")
```

tests/testthat/test-wb_functions.R
olivroy, line 37:
```suggestion
  expect_error(convertToExcelDate(df = exp["Var5"], date1904 = FALSE), "defunct")
```

JanMarvin:
Thank you! 🎉

olivroy:
I just ran the revdep check workflow just to make sure it is fine (I didn't see anything, but may still want to make sure!)

JanMarvin:
Good thinking!

JanMarvin:
Just for a follow up: we might want to either add some tests for the `deprecated.R` file or skip this file in [coverage](https://app.codecov.io/gh/JanMarvin/openxlsx2/blob/main/R%2Fdeprecate.R) testing. Usually I would prefer testing it, but since these are all deprecated functions, skipping should not hurt (`#nocov begin` and `#nocov end`, iirc).

olivroy:
You are right. or add the file to `.covrignore`! I don't see much value in testing them. FWIW, I tested them while sending the PR since some tests were still using them (and they still worked!)

JanMarvin:
I have no doubts, it just looks a little red in the coverage chart :smile:

PR #1065, opened 2024-06-28
Update .covrignore

olivroy:
Don't test deprecated functions https://github.com/JanMarvin/openxlsx2/pull/1064#issuecomment-2197331975


JanMarvin (approved):
Thanks!

Issue #1072, opened 2024-07-04
return shared formulas


d-lukasz:
Hi!
I really like the package, helps me a lot when working with .xlsb files. Now I wanted to use the function wb_to_df to compare the formulas between the two files. I can see than whenever the formula was dragged down in the column, the formula is displayed only twice then nothing. I understand this is an approach to compress the information. When doing wb_load I can see an object being created table$worksheets[[5]]$sheet_data$cc that stores information on formula and if it "shared".
Would it be possible to add the feature that allows to display all the formulas, even if they are dragged? Or maybe there is some easy way to override this that I can implement on my side?

JanMarvin:
Hi @d-lukasz , thanks for the kind words! I’ve been working a lot on xlsb lately and I’m always happy to get some feedback. Especially since I don’t use it a lot.

regarding shared formulas: unfortunately I don’t have the answer you are looking for. We’d have to modify the formulas from the original cell. Unfortunately this is not trivial for all formulas. Especially if there are fixed columns or rows. If you want to look into this, that would be a very welcome addition.

The shared cells are in the `cc` table `f_si` column. Iirc it’s a zero based index.

JanMarvin:
If you do not care about the formulas being resolved, you can try something like this:

``` r
library(openxlsx2)
library(dplyr)

packageVersion("openxlsx2")
#> [1] '1.7.0.9000'

# load an example file
wb <- wb_load("https://github.com/JanMarvin/openxlsx-data/raw/main/formula_checks.xlsb")

# shared formulas are on Sheet4 in range A9:D12
wb_to_df(wb, sheet = "Sheet4", dims = "A9:D12", col_names = FALSE)
#>             A          B          C          D
#> 9  2024-06-27 2024-06-27 2024-06-27 2024-06-27
#> 10 2024-06-27 2024-06-27 2024-06-27 2024-06-27
#> 11 2024-06-27 2024-06-27 2024-06-27 2024-06-27
#> 12 2024-06-27 2024-06-27 2024-06-27 2024-06-27

# only the formula that is shared returns a formula
wb_to_df(wb, sheet = "Sheet4", dims = "A9:D12", col_names = FALSE, show_formula = TRUE)
#>             A          B          C          D
#> 9     TODAY() 2024-06-27 2024-06-27 2024-06-27
#> 10 2024-06-27 2024-06-27 2024-06-27 2024-06-27
[...]

PR #1074, opened 2024-07-08
[fml] create shared formulas

JanMarvin:
``` r
library(openxlsx2)

df <- data.frame(
  x = 1:5,
  y = 1:5 * 2
)

wb <-  wb_workbook()$add_worksheet()$add_data(x = df)

wb$add_formula(
  x      = "=A2/B2",
  dims   = "C2:C6",
  shared = TRUE
)

wb$add_formula(
  x      = "=A$2/B$2",
  dims   = "D2:D6",
  shared = TRUE
)

if (interactive()) wb$open()
```

### DIFF
--- a/R/write.R
+++ b/R/write.R
@@ -467,12 +467,19 @@ write_data2 <- function(
     ## all have shared and all share the same f_si
     ## only the reference cell has a formula
     ## only the reference cell has the formula reference
+
+    uni_si <- unique(wb$worksheets[[sheetno]]$sheet_data$cc$f_si)
+    int_si <- as.integer(
+      replace(
+        uni_si,
+        uni_si == "",
+        "-1"
+      )
+    )
+
     cc$f_t              <- "shared"
     cc[1, "f_ref"]      <- dims
     cc[2:nrow(cc), "f"] <- ""
-
-    int_si <- as.integer(wb$worksheets[[sheetno]]$sheet_data$cc$f_si)
-    int_si[is.na(int_si)] <- -1L
     cc$f_si             <- max(int_si) + 1L
   }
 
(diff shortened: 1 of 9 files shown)


JanMarvin:
My inspiration was this [SO](https://stackoverflow.com/a/78718986/12340029). Maybe you can double check @olivroy ? Just to make sure that nothing unintended happens.

This is somewhat related to #1072

olivroy:
I can check more thoroughly, but may be worth adding an example of this somewhere?

R/write.R
```diff
@@ -459,6 +461,21 @@ write_data2 <- function(
...
+    # This cc contains only the formula range.
+    ## the top left cell is the reference
+    ## all have shared and all share the same f_si
+    ## only the reference cell has a formula
+    ## only the reference cell has the formula reference
+    cc$f_t              <- "shared"
+    cc[1, "f_ref"]      <- dims
+    cc[2:nrow(cc), "f"] <- ""
+
+    int_si <- as.integer(wb$worksheets[[sheetno]]$sheet_data$cc$f_si)
```
JanMarvin, line 474: Use unique

JanMarvin:
You always think in the bigger picture 👍  

😄 My line of thought was something like this
* hm this SO does not work
* hm I can make this work
* hm maybe this can be solved with shared strings
* oh my tiny PR looks like it's working
* darn, does not work
* why doesn't it work!
* ah, if I exclude this, it might work
* what is this `enforce` argument and why isn't it documented?
* gnar I have to fix this
* oh it works!
* Lets push to github, maybe Oliver knows what kind of checks are still missing

olivroy:
Don't forget to increment version number too!

PR #682, opened 2023-07-11
Update vignettes to snake_case, and other code quality updates

olivroy:
After converting arguments to snake case in #678, this PR updates some of the documentation, along with other code quality improvements.

Addresses #681, #643, #38, follow-up to #678

You may be best suited to review the reference index in `_pkgdown.yml`. before this PR, functions are listed twice. You made a great job of removing exports or adding `@keywords internal` in some places to reduce the number of exported functions to simplify the function index.

This PR may be a WIP, but may need your views on some topics.

Still not done with reviewing vignettes, not sure how thorough we want to be with row, cols -> dims.

The examples are not done yet.
I didn't have time to go through all vignettes.

When reviewing, please take a look at the rendered .Rd files to make sure nothing got messed up when I removed tags

* Moving test-related files out of `R/` in `tests/testthat/helper.R` See https://r-pkgs.org/testing-design.html#sec-testing-design-tension
* Revisit vignettes (rename intro, so it appears as Get Started on the pkgdown website.
* Use dims
* fix typos
* Tweak slightly some vignettes
* Revisit function index
* Separate package options in its own .Rd file.
* Fix DESCRIPTION
* Remove `@keywords internal` for internal un-exported functions.

For ref, I didn't encounter breakages or R 4.2.3, on Windows!

Edit 2: about `@keywords internal` vs `@noRd`

the `@keywords internal` I removed were related to unexported functions that already hada `@noRd` tag

Basically,

`@keywords internal`: For exported functions where you don't want the function to show up in the reference index, but still want accessible for thr help `?function` (i.e. still create the Rd file)

`@noRd` Usually For non-exported functions with roxygen documentation, this ensures that no .Rd file is created. therefore `@keywords internal` is redundant.

I did this cleanup to better represent which functions are exported, but don't show up in the ref index (i.e are not part of the suggested API, or deprecated) but still exported to avoid code breakage.

PR #1076, opened 2024-07-08
Add dark mode to website

olivroy:
Addresses a wish in https://github.com/JanMarvin/openxlsx2/pull/682#discussion_r1260262193

Now supported with [pkgdown 2.1](https://www.tidyverse.org/blog/2024/07/pkgdown-2-1-0/)

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -9,6 +9,7 @@ home:
 
 template:
   bootstrap: 5
+  light-switch: true
 
 development:
   mode: auto


JanMarvin (approved):
Lovely!

JanMarvin:
Colors look a bit strange, but I probably will get used to this. Do you know if something similar can be done for the book? (I've added bits on data validation recently, vml is still lacking documentation, but it's getting closer to be a functional something ...)

olivroy:
Feel free to open an issue in pkgdown! As for Quarto, the documentation says the following https://quarto.org/docs/output-formats/html-themes.html#dark-mode

PR #1090, opened 2024-07-20
[writing] provide openxlsx2.export_with_pugi = FALSE option

JanMarvin:
This works on my linux machine with 32gb of memory, whereas it ~~fails to export with pugixml due to the dom blowing up~~ needs approximately 6gb more with pugixml.

``` r
library(openxlsx2)

n <- 10000
k <- 1000
mm <- matrix(rnorm(n * k), ncol = n, nrow = k)

temp <- temp_xlsx()

options("openxlsx2.export_with_pugi" = FALSE)
write_xlsx(x = mm, temp)
```

Interestingly reading does not requires that much memory.

```r
wb <- openxlsx2::wb_load(temp)
```

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -348,6 +348,10 @@ set_sst <- function(sharedStrings) {
     .Call(`_openxlsx2_set_sst`, sharedStrings)
 }
 
+write_worksheet_slim <- function(sheet_data, prior, post, fl) {
+    invisible(.Call(`_openxlsx2_write_worksheet_slim`, sheet_data, prior, post, fl))
+}
+
 write_worksheet <- function(prior, post, sheet_data) {
     .Call(`_openxlsx2_write_worksheet`, prior, post, sheet_data)
 }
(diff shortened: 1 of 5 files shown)


JanMarvin:
Unfortunately this isn't the magic key to really using way less memory. We create quite a few similar sized data objects, as in similar sized to `x`. Hence the `wb_add_data()` object is consuming that much memory, that certain larger objects cannot even be written into the wb. Therefore they also cannot be exported. Presumably it is possible to shrink some of the objects (`rtype` can be removed earlier, do we create a larg dims?), but I'm not sure this will bring much. And also it isn't that much fun to program and wait until the CPU fills all memory to crash.

JanMarvin:
Just to add another thought: in theory we should be able to shrink our `cc` object and move e.g. all formula related columns into a smaller sized data frame. This would reduce the size of `cc` a lot, especially in cases where we have to write a lot of data. Currently we create a row of roughly eight values for each cell in the data.

JanMarvin:
Closing for now. I might come back to this, but at the moment it does not seem all that useful to me.

PR #1091, opened 2024-07-23
[read] read shared formulas. closes #1072

### LINKED ISSUE #1072: return shared formulas

JanMarvin:
First attempt to fix #1072


JanMarvin:
it might come in handy one day, to move this into a custom function to alter `cc` in the workbook, because IIRC there is an issue, if we write onto a cell that is part of a shared formula, hence if this works, we should be able to convert the shared formula cell into a non shared formula cell first, that we can afterwards write over safely.

PR #1094, opened 2024-07-30
[wb_dims] without `x` and non sequential rows, incorrect results were returned. fixes #1093

### LINKED ISSUE #1093: Inconsistent `wb_dims` return when row vector contains 1
`wb_dims` gives weird dimension range when the `rows` argument is a vector and the vector contains 1.

```r
r$> packageVersion("openxlsx2")
[1] '1.8.0.9000'

r$> wb_dims(rows = c(2,30), cols = 2)
[1] "B2:B2,B30:B30"

r$> wb_dims(rows = c(1,30), cols = 2)
[1] "B1:B2"

r$> wb_dims(rows = c(2,30), cols = c(2,4))
[1] "B2:B2,D2:D2,B30:B30,D30:D30"

r$> wb_dims(rows = c(1,30), cols = c(2,4))
[1] "B1:B2,D1:D2"

r$> wb_dims(rows = c(2,1), cols = c(2,4))
[1] "B1:B2,D1:D2"

r$> wb_dims(rows = c(5,1), cols = c(2,4))
[1] "B1:B2,D1:D2"
```

JanMarvin:
The output still returns ranges, but that's acceptable. We should be able to catch this and should be able to return `A1:A1` as `A1`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,10 @@
 
 * Experimental support for reading shared formulas. If `show_formula` is used with `wb_to_df()`, we try to show the value that is shown in spreadsheet software. [1091](https://github.com/JanMarvin/openxlsx2/pull/1091)
 
+## Fixes
+
+* Previously, if only `cols` and `rows` were passed to `wb_dims()` incorrect results were returned. This has been fixed. [1094](https://github.com/JanMarvin/openxlsx2/pull/1094)
+
 
 ***************************************************************************
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
 
 ## Fixes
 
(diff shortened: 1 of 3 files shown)


JanMarvin:
Maybe this should be included into the check a few lines above. There are a few things going on and obviously something unhandled slipped through and this PR feels a little hackish.

JanMarvin:
It's better now that everything is in one condition, but I'm still surprised that crossing row 1 is such an issue. It's still hackish and mysteries in my code are not pleasing 😄

olivroy (approved):
LGTM! Agreed that this is strange..

JanMarvin:
Thanks! Rebased on main.

PR #1106, opened 2024-08-14
Avoid adding the worksheet styling image on the CRAN vignette

olivroy:
I noticed that this image alone had 500 kb. 

Here is an approachto include this image only on pkgdown, but not on CRAN.

The cf images are pretty small around 30kb on average.


JanMarvin:
Thanks! Good idea, should have looked at individual file size a long time ago 🙈 
I’ll review later today

JanMarvin (approved):
Very cool! Thanks a lot!

Issue #1104, opened 2024-08-13
`wb_dims()` with `above` seems incorrect?


olivroy:
Just got to try `above`, `below` arguments. Sorry for being slow.

I think there may be an issue with `above`, unless I am misunderstanding. Will try to update the docs if I am mistaken..

```r
wb_dims(from_dims = "A3:I3", above = 1)
#> "A2"
```

I would expect this to be "A2:I2"?

```r
wb_dims(x = mtcars, from_dims = "A3", above = 1)
#> Warning:
#>  rows cannot be above of row 1 (integer position 1). resetting
#> "A1:K33"
```

I would expect this to be "A2:K34"

```r
wb_dims(x = mtcars, from_dims = "A3", select = "col_names", above = 1)
#>   rows cannot be above of row 1 (integer position 1). resetting
"A1:K1"
```

I would expect this to be "A2:K2". Let me know what you think?

https://github.com/JanMarvin/openxlsx2/pull/960

JanMarvin:
Thanks for looking into this!

## Example 1
> ```r
> wb_dims(from_dims = "A3:I3", above = 1)
> #> "A2"
> ```
> 
> I would expect this to be "A2:I2"?

I guess it is fine. We do not know the size of the element that is supposed to be placed there. If we use this cell `"A2"`, to write data, this would work, because we currently are not checking, if the object fits into `dims` and simply increase `dims` until it fits.

We could pass an object, to get the wanted size, but that's another call:
```R
# we have to use the t() trick, because otherwise it is assumed that vectors require vertical space
wb_dims(from_dims = "A3:I3", above = 1,  x = t(1:10), col_names = FALSE)
#> "A2:I2"
```

Just to confirm: the other arguments provide the same single cell in this case:

```r
> wb_dims(from_dims = "A3:I3", below = 1)
#> "A4"
> wb_dims(from_dims = "A3:I3", right = 1)
#> "J3"
> wb_dims(from_dims = "A3:I3", left = 1) # with a warning
#> "A3"
```

## Example 2
> ```r
> wb_dims(x = mtcars, from_dims = "A3", above = 1)
> #> Warning:
> #>  rows cannot be above of row 1 (integer position 1). resetting
> #> "A1:K33"
> ```
> 
> I would expect this to be "A2:K34"

[...]

PR #1107, opened 2024-08-15
Wb dims bug

olivroy:
Attempt to address partially #1104, by getting rid of some bugs?

This doesn't really work.

Previously, there was a large condition that would just evaluate the `below` `above`, `left`, `right` code if `from_dims` was supplied.

This doesn't yet work, but I think it is a start.

### DIFF
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -127,14 +127,46 @@ test_that("wb_dims() works when not supplying `x`.", {
 
 test_that("wb_dims() works with above and one line", {
   expect_equal(wb_dims(x = names(mtcars), from_row = 2), "A2:A12")
+
   # doesn't work
   # expect_equal(wb_dims(x = names(mtcars), from_row = 3, above = 1), "A2:A12")
-  expect_equal(wb_dims(x = names(mtcars), from_row = 2, above = 1), "A1:A11")
-  expect_equal(wb_dims(x = names(mtcars), from_dims = "A2", above = 1), "A1:A11")
+
+  expect_warning(
+    expect_equal(
+      wb_dims(x = names(mtcars), from_row = 2, above = 1),
+      "A1:A11"
+    ),
+    "rows cannot be above of row 1 "
+  )
+
+  expect_warning(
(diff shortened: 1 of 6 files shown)


JanMarvin (approved):
I’ve added a few minor changes. Looks good otherwise!

tests/testthat/test-utils.R
```diff
@@ -125,6 +125,50 @@ test_that("wb_dims() works when not supplying `x`.", {
...
+    ),
+    "rows cannot be above of row 1 "
+  )
+
+  expect_warning(
+    expect_equal(
+      wb_dims(x = names(mtcars), from_dims = "A2", above = 1),
+      "A1:A11"),
+    "rows cannot be above of row 1 "
+  )
```
olivroy, line 147: I still don't understand this :( 


I feel like if I read this literally. Let's use a simple example. 

```r
wb_dims(from_row = 3)
#> "A3"
# I feel like reading this literally, we just want the cell above.
wb_dims(from_row = 3, above = 1)
# I would expect this to be "A2"
```

Sorry for being slow here.

JanMarvin: The vector size is `length(names(mtcars))`. Above 1 tries to add this vector on top of `“A2“`. But there’s only place for a single value. Above, below, left and right simply don’t move integer rows in any direction, but in this case `dim(x)` wise. I tried to make it easier to move around, without having to add the object size multiple times.

If the integer passed to above is greater than 1, the distance between from_dim and the new dim object will be 1 row. I can create a more detailed example later. I guess it’s easier to understand, if you have enough space to the left and right. Maybe start in `"M40"`.


You are simply expecting a different behavior. I didn’t want to move the dims region row or column wise over the sheet.

JanMarvin: Haha, I hope you tell me if it makes no sense. I have added an example to discussion #1108 , hopefully this makes it clearer what I was intending. What I was striving for was minimizing position calculations. Because otherwise every time when the size of a data frame changes there was a chance of the calculation to be wrong.

PR #1111, opened 2024-08-18
[write] skip blanks

JanMarvin:
This PR skips blank cells entirely as discussed in #1110 .

This reduces the file size significantly for the example data provided and other sparse matrices that are thrown at `openxlsx2`, but due to possible faulty impacts, this is still unlikely to be added without extensive testing.

Maybe merge after the next release or never.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # openxlsx2 (development version)
 
+## New features
+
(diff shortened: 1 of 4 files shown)


JanMarvin:
We've reached number 1111 🥳

JanMarvin:
This needs some testing with third party software like power query and the like

JanMarvin:
@Jeff-White-AZ I've merged this with the main development branch. I'm still a bit nervous about unintended side effects like broken tables, but until it ships in a release, there's probably no way to test this fully.

PR #1119, opened 2024-08-23
[misc] document compression level option. closes #1113

### LINKED ISSUE #1113: Add compression level to options
https://github.com/JanMarvin/openxlsx2/blob/7690c3e2cc1229f532c62cd9d8233184e474b322/R/class-workbook.R#L3333

JanMarvin:
Could you please double check @olivroy , I fixed a typo in `compressionlevel` (previously with three `s`, therefore no `l`). If you have a smarter solution at hand, I'm all ears :)

This looks kinda strange in comparison
```R
getOption("openxlsx2.compressionlevel", getOption("openxlsx2.compresssionevel", 6L))
```

[Edit] updated for the missing `l`

### DIFF
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -3325,12 +3325,17 @@ wbWorkbook <- R6::R6Class(
       # TODO make self$vbaProject be TRUE/FALSE
       tmpFile <- tempfile(tmpdir = tmpDir, fileext = if (isTRUE(self$vbaProject)) ".xlsm" else ".xlsx")
 
+      # typo until release 1.8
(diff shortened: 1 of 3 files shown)


JanMarvin:
Though, while we're at it, we could also use `snake_case`. Let me know what you think.

olivroy:
I like it better as `openxlsx2.compression_level`, but I am probably never going to use this, so feel free to disregard ;)

JanMarvin:
Hehe, obviously nobody ever did, but why not start with a clean sheet while we're at it!

PR #1125, opened 2024-08-31
[dims] simplify `wb_dims()`. fixes #1124

### LINKED ISSUE #1124: [dims] unexpected result with non continuous rows and x
``` r
library(openxlsx2)

# equal with sample cols
wb_dims(x = mtcars, cols = c(1, 4:7, 9))
#> [1] "A2:A33,D2:D33,E2:E33,F2:F33,G2:G33,I2:I33"
wb_dims(rows = seq_len(nrow(mtcars)) + 1L, cols = c(1, 4:7, 9))
#> [1] "A2:A33,D2:D33,E2:E33,F2:F33,G2:G33,I2:I33"

# unequeal with sample rows
wb_dims(x = mtcars, rows = which(mtcars$cyl == 6))
#> [1] "A2:K8"
wb_dims(rows = which(mtcars$cyl == 6) + 1L, cols = seq_along(mtcars))
#> [1] "A2:K2,A3:K3,A5:K5,A7:K7,A11:K11,A12:K12,A31:K31"
```

JanMarvin:
This breaks a few previously valid cases, where we were working around broken input. The basic idea is that we have a single dim that can be reduced with selections of columns or rows and moved around in positive ranges.

### DIFF
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -281,7 +281,7 @@ test_that("`wb_dims()` works when Supplying an object `x`.", {
   expect_equal(wb_dims(x = mtcars, rows = 5:10, from_col = "C"), "C6:M11")
   # Write without column names on top
 
-  # expect_error(wb_dims(x = mtcars, cols = 0, from_col = "C"), "You must supply positive values to `cols`")
+  expect_error(wb_dims(x = mtcars, cols = 0, from_col = "C"), "You must supply positive values to `cols`")
 
   # select rows and columns work
   expect_equal(wb_dims(x = mtcars, rows = 2:10, cols = "cyl"), "B3:B11")
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -65,8 +65,8 @@ test_that("`wb_dims()` works/errors as expected with unnamed arguments", {
     wb_dims(1:10, 1:12, from_row = 2),
     wb_dims(rows = 1:10, cols = 1:12, from_row = 2)
   )
-  expect_warning(wb_dims("1", 2))
-  expect_warning(wb_dims(from_row = "C"))
+  expect_warning(wb_dims("1", 2), "should not be a character")
+  expect_error(wb_dims(from_row = "C"), "from_col/from_row must be positive integers if supplied.")
 
   # Ambiguous / input not accepted.
   # This now fails, as it used not to work. (Use `wb_dims()`, `NULL`, or )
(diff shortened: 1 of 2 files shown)


JanMarvin:
I have tried to clean up the existing code. All these nested conditions have given me one too many headaches. Since CRAN has not yet contacted me about the build errors in the vignette, I will postpone the next release for a while. At the moment, however, I suspect that this is a problem with the vignette or the build server. So maybe it will just go away.

tests/testthat/test-utils.R
```diff
@@ -65,8 +65,8 @@ test_that("`wb_dims()` works/errors as expected with unnamed arguments", {
     wb_dims(1:10, 1:12, from_row = 2),
     wb_dims(rows = 1:10, cols = 1:12, from_row = 2)
   )
-  expect_warning(wb_dims("1", 2))
-  expect_warning(wb_dims(from_row = "C"))
+  expect_warning(wb_dims("1", 2), "should not be a character")
+  expect_error(wb_dims(from_row = "C"), "from_col/from_row must be positive integers if supplied.")
```
JanMarvin, line 69: Not sure why we should accept this. Should be save to fail.

tests/testthat/test-utils.R
```diff
@@ -111,7 +111,7 @@ test_that("wb_dims() works when not supplying `x`.", {
   expect_message(out <- wb_dims(cols = 1, 2), "Assuming the .+ `rows`")
   expect_equal(out, "A2")
   # warns when trying to pass weird things
-  expect_warning(wb_dims(rows = "BC", cols = 1), regexp = "supply an integer")
+  expect_error(wb_dims(rows = "BC", cols = 1), "`rows` is character and contains nothing that can be interpreted as number.")
```
JanMarvin, line 114: Same here

tests/testthat/test-utils.R
```diff
@@ -206,11 +206,11 @@ test_that("`wb_dims()` can select content in a nice fashion with `x`", {
 
   # Selecting a row range
   dims_row1_to_5 <- "B3:L7"
-  expect_equal(wb_dims_cars(rows = 1:5), dims_row1_to_5)
+  expect_equal(wb_dims(x = mtcars, from_row = 2, from_col = "B", rows = 1:5), dims_row1_to_5)
```
JanMarvin, line 209: I don't like these functions that are defined on top of a testthat file and then it breaks at the bottom, how am I supposed to know what is going on? 😅

tests/testthat/test-utils.R
olivroy: Agreed with the change. When I made the first amendment to `wb_dims()`, this passed and transformed it to a warning, agreed it should be an error instead

tests/testthat/test-utils.R
olivroy: Yeah! While writing the tests, I was seeing if it actually worked. I should have converted the tests to standalone pieces at the end.

olivroy (approved):
I agree with the changes in tests. Sorry for creating a monster :(

JanMarvin:
Rome was not built in a day, and I believe that we have already achieved a lot from humble beginnings. As you pointed out some time ago, I consider `wb_dims()` to be an essential part of openxlsx2.

Clearing out is always easier than creating something new! Thanks for reviewing :)

PR #1134, opened 2024-09-12
[misc] cleanup in `wb_dims` code and add curl to gha

JanMarvin:


JanMarvin:
Apparently `rcmdcheck::rcmdcheck()` is unable to locate `curl` if not listed in `DESCRIPTION`. The only way around this (that I can think of right now) without explicitly depending on `curl`, would be to install `curl` from within the helper, but I rather like `testthat` to fix their ... package.

My solution with `dns_lookup()` is slower than `curl` (if `curl` is available), but not overbearingly slow (~3ms with curl and ~28ms with base R `dns_lookup()`).

JanMarvin:
Thanks, I didn’t know that!

PR #1137, opened 2024-09-15
[WIP] begin hyperlink rewrite

JanMarvin:
1) do not remove hyperlinks from `wb$worksheet_rels`
2) do not replace `wb$worksheets[[sheet]]$hyperlinks` with hyperlink objects 
3) save what is available

* [x] fix commented tests
* [x] provide function to write native hyperlinks (~~additional PR~~)
* [ ] provide function to modify/remove hyperlinks (additional PR)
* [ ] skip adding hyperlinks to cells that already contain a hyperlink


JanMarvin:
@olivroy , I would like to know what you think about `wb_add_hyperlink()`. It's a newly introduced wrapper to create what ooxml calls “shared hyperlinks”. These “shared hyperlinks” are basically “builtins” for `=HYPERLINK()`. The hyperlink formula is what we currently use to create hyperlinks. I added `wb_add_hyperlink()` similar to `wb_add_formula()` so that the function writes the data to the cell instead of just writing an overlay.

But I'm not sure if I'm really happy with this approach, as it bundles `wb_add_data_table()` and `wb_add_data()` internally and I'm not really sure if I want that. Especially since this is related to `wb_dims()` and the input and output dims might be different.

On the pro side ... it works as expected, which is nicer than before, although at the moment the references are not yet shared. This would require identifying unique strings in the input and matching them.

olivroy:
I will have to take a deeper look, but first thing, having `x` before `dims` would be better for consistency with other methods. 

I am in favour here I think as I had difficulty using hyperlink. I remember having opened a PR adding an example on how to do it.

JanMarvin:
Thanks, it’s still work in progress. The order should be cleaned up, I agree.

olivroy:
Ok, yeah, it changes a bit the internal structure of the wbWorkbook object? Maybe it requires a clear explanation for people who may have relied on other workarounds till now to create / access hyperlinks.

JanMarvin:
As usual, I worked more on the implementation. Documentation is a secondary goal 😄

JanMarvin:
Everything should be good to go. Added documentation and NEWS entries. I'm quite happy how it has turned out. It simplifies our code, less things to stumble over, and we can provide hyperlinks that are closer to what users expect. Also something that I can remove from my invisible to-do list.
I want to have a remove and/or edit function for hyperlinks, but that can wait for another PR. This one is already way to big for my personal preferences.

olivroy:
Looks good! Great work

JanMarvin:
Thanks! There are still a few things to think about, but for now: 🎉

PR #1138, opened 2024-09-18
[image] add address argument to `wb_add_image()`

JanMarvin:
@jistria I took the liberty to convert your PR to `openxlsx2`


jistria:
Oh wow, that is very generous of you! Thank you very much for adding this feature and for letting me know! I've tried it out and it works great.

JanMarvin:
Thanks for the kind words, but you did the hard work in the `openxlsx` PR, I just applied it here :)

Also I though it might motivate you to have a look at `openxlsx2`. I like to think that I have improved the `openxlsx` logic so much that it is worth a look. For starters, adding an image to a workbook, saving it, loading it again, and adding another image works flawlessly.

jistria:
It has motivated me! I've been taking a closer look since then and am very happy to see features like `fmt_txt()`. It looks like it should be easy to make the switch!

PR #1153, opened 2024-10-06
[styles] allow updates of font elements

JanMarvin:
Related to [this SO](https://stackoverflow.com/q/79056978/12340029). If `update = TRUE` all non empty font elements will be overwritten. At the moment it is not possible to set a font element to missing. All elements are initialized with either `""` or some value. Might need some additional thinking

``` r
library(openxlsx2)

wb <- wb_workbook() |>
  wb_add_worksheet() |>
  wb_add_data(x = letters) |>
  wb_add_font(dims = wb_dims(x = letters), name = "Calibri", size = 20) |>
  # updates only the font color
  wb_add_font(dims = wb_dims(x = letters), name = "", size = "", color = wb_color("orange"), update = TRUE)

if (interactive()) wb$open()
```

### DIFF
--- a/R/class-workbook-wrappers.R
+++ b/R/class-workbook-wrappers.R
@@ -3321,6 +3321,7 @@ wb_add_fill <- function(
 #' @param shadow shadow
 #' @param extend extend
 #' @param vert_align vertical alignment
+#' @param update update
 #' @param ... ...
 #' @examples
 #'  wb <- wb_workbook() %>% wb_add_worksheet("S1") %>% wb_add_data("S1", mtcars)
@@ -3351,6 +3352,7 @@ wb_add_font <- function(
       scheme     = "",
       shadow     = "",
       vert_align = "",
+      update     = FALSE,
       ...
 ) {
   assert_workbook(wb)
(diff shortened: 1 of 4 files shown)


JanMarvin:
I'm lacking a bit the motivation to work on this. Maybe someone else wants to pick it up.

Issues I see that I wanted to work on or at least get checked prior to merging:
* [ ] Do we have to do the same for other style objects? `numfmts` only has numfmts, `fill` has various other options, what about cell styles in general?
* [ ] It's subpar that we have defaults defined for the font arguments and if we want to avoid having these updated, have to pass `""` to "trick" `update` into not using the default values. At the moment I do not see any better solution other than changing the default arguments of the function to `NULL` and do some further work on the fonts code _and_ I'm not sure I like that. How else should the function know, if the font name to use is the actual default value or the value that we want after updating.
* [ ] How does updating parts of the font styles impact [named themes](https://janmarvin.github.io/ox2-book/chapters/openxlsx2_style_manual.html#named-styles)? Are they broken afterwards, are they impacted at all?
* [ ] This wont impact colors of [`fmt_txt()`](https://janmarvin.github.io/ox2-book/chapters/openxlsx2_style_manual.html#style-strings) strings. These will remain in their own color, because the color is embedded in the XML string of the string, not of the cell format. Should we warn?

JanMarvin:
A way forward might be to accept characters for `update`. Like `update="color"` to update the font color.

PR #1166, opened 2024-10-27
[test] avoid using rnorm in checks. For some reason this is failing i…

JanMarvin:
…n a current CRAN release, but we should be able to avoid this.

From [here](https://www.r-project.org/nosvn/R.check/r-release-linux-x86_64/openxlsx2-00check.html)
```R
══ Failed tests ════════════════════════════════════════════════════════════════
  ── Failure ('test-read_from_created_wb.R:173:3'): reading with pre defined types works ──
  `got` (`actual`) not equal to `dat` (`expected`).
  
    `actual[[4]]`: 1729991700 1729991699 1729991698 1729991697 1729991696
  `expected[[4]]`: 1729988100 1729988099 1729988098 1729988097 1729988096
  
  [ FAIL 1 | WARN 0 | SKIP 69 | PASS 1624 ]
```

### DIFF
--- a/tests/testthat/test-read_from_created_wb.R
+++ b/tests/testthat/test-read_from_created_wb.R
@@ -155,12 +155,15 @@ test_that("skipEmptyCols keeps empty named columns", {
 
 test_that("reading with pre defined types works", {
 
-  dat <- data.frame(
-    numeric = rnorm(n = 5),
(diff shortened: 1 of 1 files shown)


JanMarvin:
Error went away and is not reproduceable so far. Therefore closing

PR #1200, opened 2024-12-03
[write] restore `sep` functionality. closes #1199

### LINKED ISSUE #1199: `wb_add_data()` errors when `x` contains list columns
Please correct me if I'm wrong but it looks like the `sep` argument is not implemented. As a result, `wb_add_data()` errors when `x` contains list columns. Here is an example to illustrate the issue:

``` r
library(openxlsx2)

tbl <- tibble::tibble(
  "CharCol" = LETTERS,
  "ListCol" = lapply(
    LETTERS, \(x) {
      sample(LETTERS, replace = TRUE, size = 5)
      }
  )
)

wb_workbook() |> 
  wb_add_worksheet() |> 
  wb_add_data(
    x = tbl
  )
#> Error: Not compatible with STRSXP: [type=list].

packageVersion("openxlsx2")
#> [1] '1.11.0.9000'
```

<sup>Created on 2024-12-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

JanMarvin:
Previously this was only documented but did nothing

### DIFF
--- a/src/openxlsx2_types.h
+++ b/src/openxlsx2_types.h
@@ -45,7 +45,8 @@ enum celltype {
   string_num     = 13,
   cm_formula     = 14,
   hms_time       = 15,
-  currency       = 16
+  currency       = 16,
+  list           = 17
 };
 
 // check for 1.0.8.0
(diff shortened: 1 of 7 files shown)


src/openxlsx2_types.h
```diff
@@ -55,6 +56,7 @@ enum celltype {
 #include <Rcpp.h>
 #endif
 
+
```
JanMarvin, line 59: unintended change

PR #1205, opened 2024-12-07
Add functionality to add multiple sparklines as a group.

trekonom:
Inspired by this [SO](https://stackoverflow.com/questions/78522719/openxlsx2-ability-to-create-speaklines-in-a-group/78523155#78523155) post (where @JanMarvin already provided a workaround and I promised to provide a PR. Sorry, for the delay.) this PR extends `create_sparklines` to make it possible to create a true sparkline group by providing a cell range spanning multiple rows and columns, i.e. as desired in the SO post one can do
```
sparklines <- create_sparklines("Sheet 1", "A2:L13", "M2:M13", markers = "1")

t1 <- AirPassengers
t2 <- do.call(cbind, split(t1, cycle(t1)))
dimnames(t2) <- dimnames(.preformat.ts(t1))

wb <- wb_workbook()$
  add_worksheet("Sheet 1")$
  add_data(x = t2)$
  add_sparklines(sparklines = sparklines)
```

image

To this end the PR 

* adds two new helper functions `split_dims` and `split_dim` to split a cell range (I haven't found an option to achieve that based on available functions)
* adds a new argument `direction=` to create sparklines for either rows or columns in cases where `dims` is a cell range spanning both multiple columns and rows.
* add two tests to test the new functionality.

### DIFF
--- a/R/helper-functions.R
+++ b/R/helper-functions.R
@@ -392,6 +392,9 @@ hashPassword <- function(password) {
 #' @param dims Cell range of cells used to create the sparklines
 #' @param sqref Cell range of the destination of the sparklines.
 #' @param type Either `NULL`, `stacked` or `column`
+#' @param direction Either `row` or `col`. Should sparklines be created
+#' for rows or columns in cases where `dims` is a cell range spanning
+#' both multiple columns and rows. Defaults to `row`.
 #' @param negative negative
 #' @param display_empty_cells_as Either `gap`, `span` or `zero`
 #' @param markers markers add marker to line
@@ -419,11 +422,12 @@ hashPassword <- function(password) {
 #' @param ... additional arguments
 #' @return A string containing XML code
 #' @examples
-#' # create sparklineGroup
+#' # create multiple sparklines
 #' sparklines <- c(
 #'   create_sparklines("Sheet 1", "A3:L3", "M3", type = "column", first = "1"),
 #'   create_sparklines("Sheet 1", "A2:L2", "M2", markers = "1"),
-#'   create_sparklines("Sheet 1", "A4:L4", "M4", type = "stacked", negative = "1")
+#'   create_sparklines("Sheet 1", "A4:L4", "M4", type = "stacked", negative = "1"),
+#'   create_sparklines("Sheet 1", "A5:L5;A7:L7", "M5;M7", markers = "1")
 #' )
 #'
 #' t1 <- AirPassengers
@@ -435,12 +439,34 @@ hashPassword <- function(password) {
 #'   add_data(x = t2)$
 #'   add_sparklines(sparklines = sparklines)
 #'
+#' # create sparkline groups
+#' sparklines <- c(
+#'   create_sparklines("Sheet 2", "A2:L6;", "M2:M6", markers = "1"),
+#'   create_sparklines(
+#'     "Sheet 2", "A7:L7;A9:L9", "M7;M9", type = "stacked", negative = "1"
+#'   ),
+#'   create_sparklines(
+#'     "Sheet 2", "A8:L8;A10:L13", "M8;M10:M13",
+#'     type = "column", first = "1"
+#'    ),
+#'   create_sparklines(
+#'     "Sheet 2", "A2:L13", "A14:L14", type = "column", first = "1",
+#'     direction = "col"
+#'   )
+#' )
+#'
+#' wb <- wb$
+#'   add_worksheet("Sheet 2")$
+#'   add_data(x = t2)$
+#'   add_sparklines(sparklines = sparklines)
+#'
 #' @export
 create_sparklines <- function(
     sheet                  = current_sheet(),
     dims,
     sqref,
     type                   = NULL,
+    direction              = "row",
     negative               = NULL,
     display_empty_cells_as = "gap", # "span", "zero"
     markers                = NULL,
@@ -485,6 +511,43 @@ create_sparklines <- function(
   if (!is.null(markers) && as_xml_attr(markers) == "" && !is.null(type) && type %in% c("stacked", "column"))
     stop("markers only affect lines `type = NULL`, not stacked or column")
 
+  match.arg_wrapper(
+    direction,
+    c("row", "col"),
+    fn_name = "create_sparklines",
+    arg_name = "direction"
+  )
+
+  dims <- split_dims(dims, direction = direction)
+  sqref <- split_dim(sqref)
+
+  if (!length(dims) == length(sqref)) {
+    stop("dims and sqref must be equal length.")
+  }
+
+  sparkline <- Map(
+    function(dims, sqref) {
+      xml_node_create(
+        "x14:sparkline",
+        xml_children = c(
+          xml_node_create(
+            "xm:f",
+            xml_children = c(
+              paste0(shQuote(sheet, type = "sh"), "!", dims)
+            )
+          ),
+          xml_node_create(
+            "xm:sqref",
+            xml_children = c(
+              sqref
+            )
+          )
+        )
+      )
+    }, dims, sqref
+  )
+  sparklines <- paste(unlist(sparkline), collapse = "")
+
   sparklineGroup <- xml_node_create(
     "x14:sparklineGroup",
     xml_attributes = c(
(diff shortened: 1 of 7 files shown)


R/helper-functions.R
```diff
@@ -485,35 +548,78 @@ create_sparklines <- function(
...
+    vapply(rows, function(row) {
+      rowcol_to_dim(row = row, col = col_min)
+    }, FUN.VALUE = character(1))
+  } else {
+    vapply(cols, function(col) {
+      rowcol_to_dim(row = row_min, col = col)
+    }, FUN.VALUE = character(1))
+  }
+}
+
```
JanMarvin, line 622: My approach to this would be something like this.  Avoids the conversion to integers:
``` r
split_dims <- function(dims, direction = c("row", "col")) {
  if (is.numeric(direction)) {
    if (direction == 1) direction <- "row"
    if (direction == 2) direction <- "col"
  }
  df <- openxlsx2:::dims_to_dataframe(dims, fill = TRUE)
  direction <- match.arg(direction)
  if (direction == "row") df <- as.data.frame(t(df))
  vapply(df, FUN = function(x) {
    fst <- x[1]
    snd <- x[length(x)]
    sprintf("%s:%s", fst, snd)
  }, FUN.VALUE = NA_character_)
}

split_dim <- function(dims) {
  df <- openxlsx2:::dims_to_dataframe(dims, fill = TRUE)
  if (ncol(df) > 1 && nrow(df) > 1) 
    stop("`dims` should be a cell range of one row or one column.", call. = FALSE)
  df
}

dims <- "A1:C3"

split_dims(dims, 1)
#>       1       2       3 
#> "A1:C1" "A2:C2" "A3:C3"
split_dims(dims, 2)
#>       A       B       C 
#> "A1:A3" "B1:B3" "C1:C3"
split_dim(dims)
#> Error: `dims` should be a cell range of one row or one column.

dims <- "A1:A3"

split_dims(dims, "row")
#>       1       2       3 
#> "A1:A1" "A2:A2" "A3:A3"
split_dims(dims, "col")
#>       A 
#> "A1:A3"
split_dim(dims)
#>    A
#> 1 A1
#> 2 A2
#> 3 A3

dims <- "A1:C1"
split_dims(dims, "row")
#>       1 
#> "A1:C1"
split_dims(dims, "col")
#>       A       B       C 
#> "A1:A1" "B1:B1" "C1:C1"
split_dim(dims)
#>    A  B  C
#> 1 A1 B1 C1
```

R/helper-functions.R
```diff
@@ -485,35 +548,78 @@ create_sparklines <- function(
...
-      xml_node_create("x14:colorFirst",    xml_attributes = color_first),
-      xml_node_create("x14:colorLast",     xml_attributes = color_last),
-      xml_node_create("x14:colorHigh",     xml_attributes = color_high),
-      xml_node_create("x14:colorLow",      xml_attributes = color_low),
+      xml_node_create("x14:colorAxis", xml_attributes = color_axis),
+      xml_node_create("x14:colorMarkers", xml_attributes = color_markers),
+      xml_node_create("x14:colorFirst", xml_attributes = color_first),
+      xml_node_create("x14:colorLast", xml_attributes = color_last),
+      xml_node_create("x14:colorHigh", xml_attributes = color_high),
+      xml_node_create("x14:colorLow", xml_attributes = color_low),
```
JanMarvin, line 558: Please revert the whitespace changes, I actually like the previous code alignment.

R/helper-functions.R
```diff
@@ -418,12 +422,34 @@ hashPassword <- function(password) {
...
+#'   add_data(x = t2)$
+#'   add_sparklines(sparklines = sparklines)
+#'
 #' @export
 create_sparklines <- function(
     sheet                  = current_sheet(),
     dims,
     sqref,
     type                   = NULL,
+    direction              = "row",
```
JanMarvin, line 452: Please move this to the end (in front of the `...`). I like to minimize potentially breaking changes like changing the argument order

R/helper-functions.R
```diff
@@ -462,6 +488,43 @@ create_sparklines <- function(
...
+            "xm:sqref",
+            xml_children = c(
+              sqref
+            )
+          )
+        )
+      )
+    }, dims, sqref
+  )
+  sparklines <- paste(unlist(sparkline), collapse = "")
```
JanMarvin, line 526: I'm not sure I get why we need `Map()` here, but I haven't actually used it a lot. Have to run it and think about it

JanMarvin (changes requested):
Thanks for the follow up @trekonom ! Sometimes life gets in the way ...
I have left a few comments in my review, but I guess it's already good to go.
Please add yourself to the `AUTHORS` file and a NEW entry.

R/helper-functions.R
JanMarvin: Also shouldn't this be `NULL` by default. Otherwise this PR will break e.g.  https://github.com/JanMarvin/openxlsx2/discussions/921

JanMarvin:
This is a show and tell discussion related to this change: https://github.com/JanMarvin/openxlsx2/discussions/1029

R/helper-functions.R
trekonom: Fixed. I made sure to stick to your style. But this part slipped through.

R/helper-functions.R
trekonom: 1. Fixed. Moved the `direction` argument to the end. Concerning the second issue. 

2. Concerning the second issue. I checked the code in #921 and it works fine with the PR. The reason is that in my implementation the `direction=` argument will only have an effect it the `dims=` spans both multiple columns and rows, i.e. `split_dims` will not split the cell range spanning only one column or row (different to your proposal). 

I already was aware that this is not optimal. But based on the review I now realized that this also results in a bug in the knife-edge case where one wants to create a group of sparklines for each cell of one row or one column, e.g. `create_sparklines("Sheet 1", "A2:K2", "A34:K34", direction = "col")` will result in an error. I will have a look and propose a fix.

R/helper-functions.R
trekonom: No objections. I will adopt your suggestion.

JanMarvin:
@trekonom , I had a look and came up with the following https://github.com/JanMarvin/openxlsx2/commit/2e6aa29164508699a1de6f89a92320ed286ff071 and https://github.com/JanMarvin/openxlsx2/commit/71b354ab89e5410e79c166cec311957c33f62033 this fixes the issues, but sets `direction = NULL`. It neatly reduces my sparklines everywhere function to the following:

```R
sparklines2 <- function(x, col_names = TRUE, rows = TRUE, cols = TRUE) {
  
  col <- row <- NULL
  
  dims    <- wb_dims(x = x, select = "data")
  sqref_l <- wb_dims(x = x[0, ], from_dims = dims, below = 1)
  sqref_r <- wb_dims(x = x[, 0], col_names = FALSE, from_dims = dims, right = 1)
  
  if (rows) {
    row <- create_sparklines("Sheet 2", dims, sqref_r, direction = "row")
  }
  
  if (cols) {
    col <- create_sparklines("Sheet 2", dims, sqref_l, direction = "col")
  }
  
  c(row, col)
}
```

But I didn't want to force this on your PR, there might be better solutions.

trekonom:
Hej @JanMarvin. Sorry for the delay. Been way too busy this week. I just updated the PR. I now added your refactored versions of split_dims and split_dim. Concerning direction=NULL ... Although your suggestion works, I felt a little uncomfortable with it. :D First, the idea behind `direction="row"´ was to mimic the Excel default. More importantly, as it requires to explicitly set the `direction=` option it will result in a corrupt excel file without any warning or error message in cases where `dims` spans multiple rows or columns when the user forgets about that. To avoid that, to still have `NULL` the default and to fix the bug with the knife-edge cases in my implementation I added a second commit which - borrowing the idea from my original implementation - adds a `preserve_single` option to `split_dims`. I also updated the docs to clarify the behavior of `create_sparkline` for `direction=NULL`.

JanMarvin:
Thanks @trekonom , checked and its fine now!

Ultimately we could do better than Excel. Like detect if row or col groups are required, but we don't have to prolong this PR further. Could you rebase to `main`. I'll merge afterwards.

``` r
library(openxlsx2)

# Helper function to parse a cell range into row and column bounds
parse_range <- function(range) {
  bounds <- dimnames(openxlsx2::dims_to_dataframe(range, fill = TRUE))
  names(bounds) <- c("rows", "cols")
  bounds
}

# Function to determine if a range is row-wise, column-wise, or offset
detect_orientation <- function(reference_range, target_range) {
  ref_bounds <- parse_range(reference_range)
  tgt_bounds <- parse_range(target_range)
  
  # Check for row-wise alignment
  if (all(ref_bounds$rows %in% tgt_bounds$rows)) {
    return("rowwise")
  }
  
  # Check for column-wise alignment
  if (all(ref_bounds$cols %in% tgt_bounds$cols)) {
    return("colwise")
  }
  
  # If neither aligns, consider as offset
  return("offset")
}

# Example usage
reference      <- "A1:B2"
target_rowwise <- "A2:B2" # Same rows, horizontal extent
target_colwise <- "B1:B2" # Same cols, vertical extent
target_offset  <- "C3:D4" # Different rows and cols

cat("Colwise Test: ", detect_orientation(reference, target_rowwise), "\n")
#> Colwise Test:  colwise
cat("Rowwise Test: ", detect_orientation(reference, target_colwise), "\n")
#> Rowwise Test:  rowwise
cat("Offset Test: ", detect_orientation(reference, target_offset), "\n")
#> Offset Test:  offset
```

JanMarvin:
Oh and please add youself to `inst/AUTHORS` (either full name or github user name) and a bullet point for the `NEWS` would be lovely!

trekonom:
Just done. Rebased, added a NEWS entry and added me to authors.

JanMarvin:
Sorry, not sure what happened, but there were various duplicated commits. I rebased from the commit I accepted, added the NEWS commit and my cleanup. Now the PR looks clean.

JanMarvin (approved):

JanMarvin:
Thank you! Merged.

JanMarvin:
The comment got lost: I moved your name from DESCRIPTION to AUTHORS, because I prefer to keep the `DESCRIPTION` `Authors@R` field for long time contributors.

PR #1233, opened 2025-01-03
[gha] updates and use `covr` dev

JanMarvin:


JanMarvin:
Switch to the development version of `covr` to fix the issue reported in r-lib/covr#588 and finally enjoy normal coverage runtime 🎉

PR #1243, opened 2025-01-11
[Rcpp] add .clang-format definition and run:

JanMarvin:
`clang-format -i *.cpp`


JanMarvin:
I'm not sold yet on either style, not this one, nor #1244 
There are a few things that should be cleaned up. Especially whitespaces and blank lines, but I don't like bits and pieces in both styles.

PR #1247, opened 2025-01-13
[xlsb] fix linebreak in formulas

JanMarvin:


src/xlsb_funs.h
```diff
@@ -1168,6 +1151,14 @@ std::string parseRPN(const std::string& expression) {
     formulaStack.pop();
   }
 
+  // undo the newline escaping
+  std::string have = "&amp;#xA;", want = "\n";
```
JanMarvin, line 1155: It isn’t guaranteed that &amp; exists. Could be &#xA;

PR #1248, opened 2025-01-15
Add the missing right parenthesis

Shitao5:
Hello, openxlsx2 is really useful. While learning it, I noticed that a right parenthesis was missing here, and I've already filled it in the PR.


JanMarvin:
Thanks!

PR #1255, opened 2025-01-29
[save] Stream Worksheet XML to Disk

JanMarvin:
This is an experimental branch, using a custom C++ function to stream the XML file to disk. It is working, but not extensively tested and might still lack features. It has shown that it can reduce the memory consumption significantly.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # openxlsx2 (development version)
 
+## New features
+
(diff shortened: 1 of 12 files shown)


JanMarvin:
Cherry pick 15daf92

JanMarvin:
Using this, is a game changer memory wise. Writing a 500,000 * 100 rnorm data frame previously required - with the optimizations from this branch - ~28GB. Without pugixml using a custom xml writer, roughly 12-14GB less are required. This is also skipping a few guards that check that the cells and rows are in the correct order.

```R
options("openxlsx2.export_with_pugi" = FALSE)
```

Basically what this skips building the XML file in memory before it is written to the disk. The custom xml writer simply flushes what is available into a txt file. Without any checks for correct xml etc.

JanMarvin:
`wb_save(flush = TRUE)` is now possible

PR #1291, opened 2025-03-17
[read] check against lower case numfmt

JanMarvin:
pandas uses upper case numfmt for dates


JanMarvin:
For reference, the standard mentions the following interpreted characters:

| Interpreted Characters |
|--|
|0|
|#|
|@|
|d|
|m|
|y|
|h|
|s|
|?|
|AM/PM|
|A/P|
|g|
|e|
|r|

Issue #1213, opened 2024-12-15
[ideas][v2] thinking about API changes and deprecation warnings


JanMarvin:
This issue is to bundle some ideas for a possible v2 version.

So far `openxlsx2` avoids breaking changes if possible and there is no need to change anything yet. That's why this issue is not a, it will be implemented soon, but a could come and may take another six months or longer.

## `openxlsx2.soon_deprecated`
We still allow some deprecated usages, e.g. entering `rows`/`cols` in various functions. It is probably time that we enable the `openxlsx2.soon_deprecated` warning. We should deliver this warning for some time, maybe six months, and then remove the old behavior (and the warning).

## `standardize()`
Although the addition of `standardize(...)` worked well, there is a fee to pay. We do not strictly check the possible arguments passed to each function. Therefore, passing arguments to functions that are not used will not cause a warning. Therefore, fixing this problem should be at least a medium-term goal.

## write data cleanup
In addition, the functions `wb_add_data()`/`wb_add_formula()`/`wb_add_data_table()` could probably be tightened up a little. Currently, the process looks something like this:

```R
wb_add_data()          # Call the first function
-> do_write_data()     # fill write_data_table()
--> write_data_table() # differentiate between data_table() and 
---> write_data2()     # actually write the data to the worksheet
```

It is because we have been exporting the original `do_write_data()` function for a long time. And all functions perform various checks, data changes and shaping at some point. We check for hyperlinks at various stages, and until `write_data2()`, non-data frame input is allowed. This should be cleaned up as it may break the currently documented behavior.

## Arguments cleanup
For the recently added arguments, we have avoided changes in the order of the arguments. The arguments were not added in the best sensible order, but at the end of the previous arguments. Some arguments should allow different inputs or no inputs at all. See e.g. #1153

## Argument checking
We allow various unchecked inputs for many arguments. Recently I was confused about `wb_set_sheetview(view = ...)`. I used an unexpected input and was not alerted with a warning. It's probably time we introduce `match.args()` for all inputs (perhaps with an optional override option) and clean up the remaining camel case along the way.

# Input, ideas and suggestions are welcome
I am not the only one using this package, and possibly many are affected by the changes to the API. So any input on planned or desired changes is welcome, but please be reasonable. It is unlikely that `openxlsx2` will be rewritten on a large scale.

elipousson:
[...]

PR #1330, opened 2025-04-18
[misc] housekeeping deprecating

JanMarvin:
Revamped this PR entirely. Now it consists of three commits deprecating

* create_comment()
* rows/cols usage
* ~~camelCase usage~~ (reverted this for now. the text states that we will deprecate at the next major release, so we should be true to our words). Still from our internals, everything is prepared to deprecated camelCase usage

If you have a moment, could you have a look @olivroy (scrolling along the PR should be enough). This PR is touching so many places and I want to minimize the risk of something slipping through. If you are wondering, yes, there are still camelCase arguments in our internal functions and in the code, but atm I do not plan changing these.

We should add a NEWS entry, that rows/cols and camelCase conversion are deprecated/near deprecation and will not be actively tested going forward. Hence users are advised to move to snake_case usage and dims.

This is related to #1213. I'm not sure, if this is ready to go into 1.15. With the UTC change we already have one major breakage and it might be a good idea to limit these kind of changes for a real major release. On the other hand, if we start deprecating now (and no CRAN package is using create_comment(), rows/cols and just one package is using camelCase), we could defunct this with 2.0 which would be nice too.


JanMarvin:
Thanks for the review!

Issue #1346, opened 2025-05-15
Multiple non-continuous dimension in `wb_add_conditional_formatting`


J-Moravec:
I got 3 columns of different type, these columns are repeated for every sample.
I have a proper selection setup and `wb_dims` is returning the proper value.

When fed into `wb_add_conditional_formatting`, the whole range is coloured instead.

1. Technically, this is documented with a warning. But:
    - the warning didn't make sense until after I understood what is going on
    - I ended up with broken xlsx instead of failing quickly
2. This is not documented in the help pages for `wb_add_conditional_formatting`
3. This is not documented in the vignette (bless you autocorrect) for `wb_add_conditional_formatting`
4. Just for-cycling over columns solves it.

**Suggestion**
1. Add documentation in both help pages and vignette
2. Improve the error message so its obvious what is going on
3. Consider vectorising over it?

**Warning message**
```
In wb$add_conditional_formatting(sheet, dims = openxlsx2::wb_dims(rows = 2,  :
  cols > 2, will create range from min to max.
```

**MRE**
```{r}
sheet = "sheet"
wb = openxlsx2::wb_workbook()
wb = openxlsx2::wb_add_worksheet(wb, sheet)
wb = openxlsx2::wb_add_data(
    wb, sheet, matrix(1, ncol = 10, nrow = 3),
    col_names = FALSE, row_names = FALSE
    )

wb$add_dxfs_style(
    name = "red", bg_fill = openxlsx2::wb_color("red")
    )

wb$add_conditional_formatting(
    sheet,
    dims = openxlsx2::wb_dims(rows = 1, cols = 1:10),
    rule = "==1",
    style = "red"
    )

wb$add_conditional_formatting(
    sheet,
    dims = openxlsx2::wb_dims(rows = 2, cols = seq(from = 1, to = 10, by = 2)),
    rule = "==1",
    style = "red"
    )

wb$save("openxlsx2.xlsx")
```

JanMarvin:
Hi @J-Moravec ,

feel free to open a pull request. Documenting and improving the warning message should be easy, failing fast might not be and there might be conditions where it is not even wanted ... vectorization ... I'm not gonna comment further on speed issues, it triggers my PTSD 😉.

[...]

PR #1347, opened 2025-05-17
[cf] Allow multiple non-consequtive dims

JanMarvin:
This is a working implementation of #1346, but it reveals a major breakage.

Previously, the following would draw a mesh over the entire dimension, even though the `dims` argument itself would indicate indicate that only columns A and K are required. I'm not sure that I want this and probably, this would require a warning, whenever two columns are selected.

```R
df <- mtcars
wb_dims(cols = c(1, ncol(df)), rows = seq_len(nrow(df)))
#> "A1:A32,K1:K32"
```

Also this probably would require some further testing, I only tested the case below:

```R
library(openxlsx2)

set.seed(123)

  add_data(x = matrix(sample(c(0, 1), replace = TRUE, size = 100), ncol = 10, nrow = 10),
           row_names = FALSE
  )

# highlight column
wb$add_fill(dims = openxlsx2::wb_dims(rows = 1, cols = seq(from = 1, to = 10, by = 2)), color = wb_color("magenta"))

wb$add_dxfs_style(
  name = "red", bg_fill = openxlsx2::wb_color("red")
)

wb$add_conditional_formatting(
  dims = openxlsx2::wb_dims(rows = c(2, 4:5, 8), cols = seq(from = 1, to = 10, by = 2)),
  rule = "==1",
  style = "red"
)

if (interactive()) wb$open()
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@
 
 ## Breaking changes
 * `dims` will now respect the order in which they are sorted. So `"B2:A1"` will no longer equal `"A1:B2"`. [1353](https://github.com/JanMarvin/openxlsx2/pull/1353)
+* Conditional formatting allows now for non consecutive dims such as `"A1:B2,C2:D3"`. This causes a change in previous behavior, where the outer cells of a dimension were used to construct the range to which the conditional formatting was applied. [1347](https://github.com/JanMarvin/openxlsx2/pull/1347)
 
 
 ***************************************************************************
--- a/R/class-workbook-wrappers.R
+++ b/R/class-workbook-wrappers.R
@@ -4063,6 +4063,7 @@ wb_add_form_control <- function(
 #' @details
 #' Conditional formatting `type` accept different parameters. Unless noted,
 #' unlisted parameters are ignored.
+#' If an expression is pointing to a cell `"A1=1"`, this cell reference is fluid and not fixed like `"$A$1=1"`. It will behave similar to a formula, when `dims` is spanning multiple columns or rows (A1, A2, A3 ... in vertical direction, A1, B1, C1 ... in horizontal direction). If `dims` is a non consecutive range ("A1:B2,D1:F2"), the expression is applied to each range. For the second `dims` range it will be evaluated again as `"A1=1"`.
 #' \describe{
 #'   \item{`expression`}{
 #'     `[style]`\cr A `Style` object\cr\cr
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -5800,13 +5800,10 @@ wbWorkbook <- R6::R6Class(
         dims <- rowcol_to_dims(rows, cols)
       }
 
-      ddims <- dims_to_rowcol(dims, as_integer = TRUE)
-      rows <- ddims[["row"]]
-      cols <- ddims[["col"]]
-
-      if (length(cols) > 2 && any(diff(cols) != 1)) {
-        warning("cols > 2, will create range from min to max.")
-      }
+      # as_integer returns a range, but we want to know all columns
+      ddims <- dims_to_rowcol(dims, as_integer = FALSE)
+      rows <- as.integer(ddims[["row"]])
+      cols <- col2int(ddims[["col"]])
 
       if (!is.null(style)) assert_class(style, "character")
       assert_class(type, "character")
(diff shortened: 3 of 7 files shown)


JanMarvin:
@J-Moravec could you give this one a try, sometime this week? I'm not sure yet what to do with the potentially breaking case. There are multiple conditional formatting cases and I'm not sure that this is a one size fits all solution. As in, it might work for the test case above, but not for the remaining cases.

J-Moravec:
Sorry, busy week. I will test it tomorrow.

J-Moravec:
Looks good so far.

For equality, the XML content looks like his:

```
<conditionalFormatting sqref="A2:C3">
<cfRule type="expression" dxfId="0" priority="3">
<formula>A2=1</formula>
</cfRule>
</conditionalFormatting>
<conditionalFormatting sqref="E2:E3">
<cfRule type="expression" dxfId="0" priority="2">
<formula>E2=1</formula>
</cfRule>
</conditionalFormatting>
<conditionalFormatting sqref="H2:J3">
<cfRule type="expression" dxfId="0" priority="1">
<formula>H2=1</formula>
</cfRule>
```
The "formula" seems strange, but looks like some kind of stuff that is reinterpreted based on cell of interest (at least in LibreOffice) and doesn't really depends on e.g., H2=1 for the whole H2:J3.

Tested it with "between" rule and it works as well.

The only way I was able to break it is:

1. Columns provided in reverse order e.g, 3:1 instead of 1:3 creates a comma-separated columns instead of single range in `wb_dims`.
2. Fix the cell in formula, e.g., `A1=1`. Consider following example:

```r
sheet = "sheet"
wb = openxlsx2::wb_workbook()
wb = openxlsx2::wb_add_worksheet(wb, sheet)
wb = openxlsx2::wb_add_data(
    wb, sheet, matrix(1, ncol = 10, nrow = 3),
    col_names = FALSE, row_names = FALSE
    )

wb$add_dxfs_style(
    name = "red", bg_fill = openxlsx2::wb_color("red")
    )


wb$add_conditional_formatting(
    sheet,
    dims = openxlsx2::wb_dims(rows = c(1:3), cols = c(1:3, 8:10)),
    rule = "A1=1",
    style = "red"
    )

wb$save("openxlsx2.xlsx")
```

change the dims to: `wb_dims(rows = 3:1, cols = c(1:3, 8:10)`, `wb_dims(rows = 3:1, cols = c(1:3, 10:8)`, `wb_dims(rows = 3:1, cols = c(3:1, 10:8)` to get various forms of strange behaviour.

I don't think this is necessarily a bug, just a strange way how LibreOffice (at least) interprets formulas:

When defined as:
```
<conditionalFormatting sqref="A1:C3">
<cfRule type="expression" dxfId="0" priority="2">
<formula>A1=1</formula>
</cfRule>
</conditionalFormatting>
<conditionalFormatting sqref="H1:J3">
<cfRule type="expression" dxfId="0" priority="1">
<formula>A1=1</formula>
</cfRule>
</conditionalFormatting>
```
In both cases, the A1 is expanded depending on the range of the region. But it always starts with A1, so given both regions have 3 rows and 3 columns, the A1 is expanded to A1:C3.

But if the columns are provided in reverse order to wb_dim, ranges are not constructed and we get some funky behaviour.

Again, I don't think this is a bug, but maybe there might be a place in documentation about it? (I love when tools are also teaching materials about the topic)

JanMarvin:
Thanks! Regarding the following:
``` r
library(openxlsx2)

wb_dims(rows = 3:1, cols = c(1:3, 8:10))
#> [1] "A3,B3,C3,H3,I3,J3,A2,B2,C2,H2,I2,J2,A1,B1,C1,H1,I1,J1"
wb_dims(rows = 3:1, cols = c(1:3, 10:8))
#> [1] "A3,B3,C3,J3,I3,H3,A2,B2,C2,J2,I2,H2,A1,B1,C1,J1,I1,H1"
wb_dims(rows = 3:1, cols = c(3:1, 10:8))
#> [1] "C3,B3,A3,J3,I3,H3,C2,B2,A2,J2,I2,H2,C1,B1,A1,J1,I1,H1"
```

I agree that it looks strange, I'll try to address this in another PR. Probably we check something like `diff(rows) != 1` when we could check `abs(diff(rows)) != 1`.

JanMarvin:
I have tweaked the behavior of `dims` a bit. So that decreasing columns and rows are possible. Still, your example of conditional formatting above is not trivial to understand.

```R
wb$add_conditional_formatting(
  sheet,
  dims = openxlsx2::wb_dims(rows = c(3:1), cols = c(3:1, 8:10)),
  rule = "A1=1",
  style = "red"
)
```

With this PR the code above is broken down into various pieces from either this:
```R
openxlsx2::wb_dims(rows = c(3:1), cols = c(3:1, 8:10))
#> [1] "C3:C1,B3:B1,A3:A1,H3:H1,I3:I1,J3:J1"
```
or from the shorter form
```R
#> [1] "C3:A1,H3:J1"
```

The loop will move over rows 3, 2, 1 and columns 3, 2, 1, 8 - 10. Because of this, many segments of the resulting conditional formatting will consist of a single cell (C3, B3, A3 etc) and a few column spanning conditions, namely: H3:J3, H2:J2, H1:J1. For each segment (single cells and ranges) the condition A1=1 is checked.

In conditional formatting, `A1=1` is unequal to `$A$1=1`. So spreadsheet software will evaluate `A1=1` for every single cell segment and for the first cell of a range. For the second horizontal cell of a range `A2=1` is evaluated, for the third horizontal cell `A3=1`.

Therefore, while conditional formatting is applied and working, it causes bad surprises. The conditional formatting very likely wont work as intended

JanMarvin:
This should be fixed now. In the example below (yours from above, just a bit cleaned up), the two columns on the outer borders behave exactly the same.

``` r
library(openxlsx2)
wb <- wb_workbook()$
  add_worksheet("sheet")$
  add_data(
    x = matrix(1, ncol = 10, nrow = 3),
    col_names = FALSE, row_names = FALSE
  )

wb$add_dxfs_style(
  name = "red", bg_fill = wb_color("red")
)


wb$add_conditional_formatting(
  dims = wb_dims(rows = c(3:1), cols = c(3:1, 8:10)),
  rule = "A1=1",
  style = "red"
)

# wb$add_conditional_formatting(
#   dims = "C3:A1,H3:J1",
#   rule = "A1=1",
#   style = "red"
# )
# wb$worksheets[[1]]$conditionalFormatting

if (interactive()) wb$open()
```

JanMarvin:
Thanks for the test @J-Moravec , it helped identify issues and I believe the PR now in a way better shape than with my initial draft. I merge it

PR #1353, opened 2025-05-23
[dims] handle decreasing consecutive rows and cols

JanMarvin:
addresses https://github.com/JanMarvin/openxlsx2/pull/1347#issuecomment-2902568052


JanMarvin:
rows do not decrease:

```R
wb_dims(rows = 3:1, cols = 1:3)
#> [1] "A1:C3"
```

PR #1397, opened 2025-07-11
[cf] support internal convert conditional formatting to data frame

### LINKED ISSUE #1396: pivot table conditional formatting
Hey, thanks so much for all your work on the package, it's been incredibly useful.

I was trying to load a template containing a pivot table with multiple conditional formatting rules applied (simple example attached) and was getting an error:

``` r
template <- "pivot_multiple_conditional_formatting.xlsx"
wb <- openxlsx2::wb_load(template)
#> Error in names(cf) <- nms: 'names' attribute [4] must be the same length as the vector [2]
```

<sup>Created on 2025-07-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

After playing around with `wb_load()` and `class-worksheet.R` I was able to get this template to load and maintain the conditional formatting on the pivot table.

If you thought this might be useful then I'm happy to raise a PR for you to take a look? Or I can copy code here, the changes were pretty small.

Thanks!

[pivot_multiple_conditional_formatting.xlsx](https://github.com/user-attachments/files/21184775/pivot_multiple_conditional_formatting.xlsx)

JanMarvin:
closes #1396

* this converts the default conditional formatting to a data frame, so that the pivot attribute can remain. 
* the x14 attributes go to another argument. these were probably handled correct.
* removing a conditional formatting for a pivot or removing a pivot without the conditional formatting might brick the spreadsheet

### DIFF
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -6152,15 +6152,23 @@ wbWorkbook <- R6::R6Class(
 
         cf <- self$worksheets[[sheet]]$conditionalFormatting
 
-        if (!is.null(dims)) {
-          if (any(sel <- names(cf) %in% dims)) {
-            self$worksheets[[sheet]]$conditionalFormatting <- cf[!sel]
+        if (is.data.frame(cf)) {
+
+          if (!is.null(dims)) {
+            if (any(sel <- cf$sqref %in% dims)) {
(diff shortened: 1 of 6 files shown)


JanMarvin:
Not sure what happend with the commit message :sweat_smile:

Issue #1408, opened 2025-07-26
Consider removing `magrittr` from Imports


JanMarvin:
We import the pipe operator `%>%` from the `magrittr` package. It appears that the `magrittr` package is no longer actively developed not even the warnings are fixed regarding illegal access of usage of the internal APIs are fixed. I don't think that CRAN will drop it anytime soon (even though I violates some internal CRAN checks), because of its many dependencies, but we probably should avoid pulling it in. Especially, because R has a native pipe operator `|>` since R 4.1.

The tricky part though, I don't want to depend on R 4.1+ just for the pipe operator. In many work environments it is not in the hands of the package users to update the R version, but in the hands of the IT department.[1] R 4.1 was release in May 2021 and it probably is a safe bet that there are many R installations around that are not yet at R 4.1+. Adding a dependency on 4.1 would actively lock them out. For something that is not even required by the package to function.

We use the pipe operator in various examples, the vignettes and tests. 
* I guess we can rewrite the tests to avoid pipe usage altogether, this is not user facing code and aside from the superfluous assurance that it works with pipes, we gain nothing
* In vignettes we could guard examples depending on `getRversion()` to use the native pipe or no pipe at all
* Similar to tests, we could probably rewrite the examples to not depend on the pipe operator as well

Any additional insights?

[1] This includes me and the outdated R installation on my $DAYJOB laptop.

PR #1409, opened 2025-07-26
[misc] remove `magrittr` from imports

JanMarvin:
#1408


JanMarvin:
This is a breaking change as in, users might have to explicitly add `library(magrittr)` to scripts that previously did not require this.

tests/testthat/test-class-comment.R
```diff
@@ -273,9 +271,8 @@ test_that("threaded comments work", {
   got <- wb_get_thread(wb, dims = "A1")[, -1]
   expect_equal(exp, got)
 
-  wb <- wb %>%
-    wb_add_worksheet() %>%
-    wb_add_thread(dims = "A1", comment = "hmpf", person_id = scotty_id)
+  wb <- wb$add_worksheet()$
```
JanMarvin, line 274:
```diff
- wb <-
```

PR #1413, opened 2025-08-04
Update NEWS.md

olivroy:


NEWS.md
```diff
@@ -1,4 +1,4 @@
-# openxlsx2 1.18
+# openxlsx2 1.19
```
JanMarvin, line 1: Oh it should be Development Version, shouldn’t it?

PR #1417, opened 2025-08-08
[dims] add type argument

### LINKED ISSUE #1416: `wb_dims()`: Change between relative, absolute and mixed
Hi. First off thank you for this package, I use it often as the output to some data analysis scripts. I came across this function and wondered whether there is any way to alter the spreadsheet range that is outputted to switch between relative, absolute or mixed references: https://support.microsoft.com/en-gb/office/switch-between-relative-absolute-and-mixed-references-dfec08cd-ae65-4f56-839e-5f0d8d0baca9 ?

For example, `wb_dims(1:4, 1:2)` returns "A1:B4". Perhaps some argument could be added e.g `wb_dims(1:4, 1:2, ref_type = "absolute")` could return "$A$1:$B$4", and different options for ref_type can output the other variants. I can see this being useful for building the formula within `wb_add_formula()`. Is it possible to implement something like this within the package? Thank you.

JanMarvin:
closes #1416

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # openxlsx2 (development version)
 
+## New features
+
+* Add `type` argument to `wb_dims()`. This allows setting the type of the dimension. The default is a relative reference, but it can be fully fixed `"all"`, rows fixed `"row"`, or columns fixed `"col"`. [1417](https://github.com/JanMarvin/openxlsx2/pull/1417)
+
 ## Breaking changes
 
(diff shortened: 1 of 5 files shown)


JanMarvin:
That was a confusing error :) Did not want to intend that you should fix the mess, @olivroy . But can you have a look? After all, `wb_dims()` is your baby and you might have an opinionated view

olivroy (approved):
Seems reasonable to me! althought the name `type` doesn’t evoke much. 

I think that adding an example would be valuable. 

I understood what this was doing by looking at the tests

PR #1466, opened 2025-11-16
[read] extend dims with + or - to max and min column or row

### LINKED ISSUE #1465: [fr][dims] read all available cells
Lately I was wondering if we could provide something like `wb$to_df(dims="A1:B1,A3:B+")` or something alike, which would be interpreted as first row and all available rows following. Similar to how start_row, works. Something alike could also be used for columns.

JanMarvin:
closes #1465

``` r
library(openxlsx2)
fl <- "https://www.cpb.nl/system/files/cpbmedia/cpb-world-trade-monitor-august-2025_0.xlsx"

wb <- wb_load(fl)

df <- wb$to_df(dims = "B4,F4:+4,B8,F8:+8", skip_empty_cols = TRUE, row_names = TRUE)

df <- as.data.frame(t(df))
df$date <- paste0(rownames(df), "d01")
df$date <- as.Date(df$date, format = "%Ym%md%d")

plot(df$date, df$`World trade`, 
     main = "CPB.nl: World Trade seasonally adjusted",
     xlab = "Months", ylab = "Volumes (USD bln, 2021 = 100)", type = "l",
     panel.first = grid())
```

### DIFF
--- a/tests/testthat/test-wb_functions.R
+++ b/tests/testthat/test-wb_functions.R
@@ -483,10 +483,10 @@ test_that("improve non consecutive dims", {
   exp <- dims_to_dataframe(dims, fill = TRUE)
   exp <- unname(unlist(exp[exp != ""]))
   got <- wb1$worksheets[[1]]$sheet_data$cc$r[wb1$worksheets[[1]]$sheet_data$cc$c_s != ""]
-  expect_contains(got, exp)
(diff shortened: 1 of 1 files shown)


JanMarvin:
This is a bit experimental, it crossed my mind two days ago on a train ride. Any objections from your side @olivroy ?

PR #1499, opened 2025-12-18
[read+write] consolidate `na.strings` and `na.numbers` into `na`

JanMarvin:
`na.strings` is a relict from openxlsx and should have been replaced with na_strings a long time ago. But it was not and now there is `na_strings()` and we cannot have `na_strings = na_strings()`. Thus this PR is the attempt to use the short form `na` instead of `na.strings`

This needs more tests. Appart from many of the previous defaults, the new `na` argument is not yet tested.

### DIFF
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -1463,7 +1463,7 @@ wbWorkbook <- R6::R6Class(
       ) {
 
       args <- list(...)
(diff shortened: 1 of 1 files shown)


olivroy (commented):
I agrree with the idea, perhaps, we should read the new option first if it exists.

Also worth adding the new option name in NEWS

PR #1508, opened 2025-12-23
[dims] remove more unreached code

JanMarvin:
Try to increase coverage once again

@olivroy FYI: I'm not entirely happy with my refactoring of `wb_dims()`,  now that the code checks for valid rows and cols arguments are spread across two functions. But this way it was easier to make sense out of it. IMHO a few of these checks are overengineered. But at least I wanted to increase the coverage for the entire `utils.R` file including `wb_dims()`.


olivroy:
I think this is great! I probably overengineered this function at the beginning.

olivroy:
You too! `wb_dims()` is not a good reason to have a sleepless night before Christmas Eve 😥

PR #1618, opened 2026-04-13
[write] escape newline in table column name

### LINKED ISSUE #1617: add_data_table() generates content error in the Excel file.
Hello,

I get an odd warning in Excel when I open a file that contains data created with `add_data_table()` and containing a newline "\n" in the name of a column.

``` r
library(openxlsx2)

# this works
wb <- wb_workbook()$
  add_worksheet("sheet1")$
  add_data(x = tibble::tibble(`line1\nline2` = c(1,2,3)))
wb_save(wb, overwrite = TRUE, file = "test-ok-openxlsx2.xlsx")

# this works in R but generate a problem in the content in Excel
wb <- wb_workbook()$
  add_worksheet("sheet1")$
  add_data_table(x = tibble::tibble(`line1\nline2` = c(1,2,3)))
wb_save(wb, overwrite = TRUE, file = "test-notok-openxlsx2.xlsx")
```

When opening "test-notok-openxlsx2.xlsx":  
Image

If I click "Yes", the file is restored and the content is as expected.

Image  

Is it a small bug or should I avoid newlines in column names ?

Thank you.

JanMarvin:
closes #1617

### DIFF
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -28,6 +28,7 @@ last_table_id <- function(wb) {
 }
 
 fun_tab_cols <- function(tab_cols) {
+  tab_cols <- escape_newline_and_tab(tab_cols)
   tabCols <- NULL
   for (i in seq_along(tab_cols)) {
     tmp <- xml_node_create(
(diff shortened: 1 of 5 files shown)


JanMarvin:
Add this to update_table as well

Issue #1640, opened 2026-06-15
wb_load() + wb_save() round-trip drops vmlDrawing.vml.rels and leaves a dangling pageSetup r:id


SchmidtPaul:
A bare `wb_load()` + `wb_save()` round-trip (no modification at all) of a
structurally richer workbook can produce a file that Excel opens with its
"We found a problem with some content" repair prompt. Two independent causes,
both still present on the dev HEAD (`1.27.0.9000`) and on CRAN `1.27`:

* **A - `vmlDrawing*.vml.rels` loss** (regression in 1.24)
* **B - dangling `pageSetup r:id`** for dropped `printerSettings` (since &ge; 1.0)

I have a one-line fix + regression test ready for the first one (PR linked
below). The second is more of a design question for you - details at the end.

## A. vmlDrawing.vml.rels dropped on round-trip

Self-contained reprex (uses the bundled `einstein.jpg`, so no external file):

```r
library(openxlsx2)
img <- system.file("extdata", "einstein.jpg", package = "openxlsx2")

# three sheets, each with a comment carrying a background image
# -> three xl/drawings/_rels/vmlDrawing*.vml.rels
wb <- wb_workbook()
for (i in 1:3) {
  wb$add_worksheet()
  wb$add_comment(sheet = i, dims = "B2",
                 comment = wb_comment(text = "x", author = "x"), file = img)
}

count_vmlrels <- function(f)
  sum(grepl("vmlDrawing.*\\.vml\\.rels$", unzip(f, list = TRUE)$Name))

f1 <- temp_xlsx(); wb$save(f1)
count_vmlrels(f1)
#> [1] 3

# pure load + save, nothing else
f2 <- temp_xlsx(); wb_load(f1)$save(f2)
count_vmlrels(f2)
#> [1] 1   # two .vml.rels dropped; their o:relid refs now dangle

packageVersion("openxlsx2")
#> '1.27.0.9000'
```

Root cause is in `wb_load()`: `wb$vml_rels <- rep(list(""), vml_len)` sits
*inside* the `for (vml_rel in vmlDrawingRelsXML)` loop, so it is wiped on every
iteration and only the last sheet's entry survives. The sibling `wb$vml` loop a
few lines above initialises *before* its loop, which is the correct pattern.
This was introduced in #1557 (shipped in 1.24), which deferred the allocation -
the right idea, the line just landed one block too deep.

Fix is to move the (length-guarded) init out of the loop. PR: #1641 .

[...]

PR #1641, opened 2026-06-15
[load] fix wb$vml_rels re-init inside loop dropping all but last .vml.rels

SchmidtPaul:
Fixes the `vmlDrawing*.vml.rels` half of #1640 .

`wb_load()` re-initialised `wb$vml_rels` on every iteration of the
`vmlDrawingRelsXML` loop, so only the last worksheet's `vmlDrawing*.vml.rels`
survived a `wb_load()` + `wb_save()` round-trip. The other entries were dropped
and their `o:relid` references left dangling, which triggers Excel's repair
prompt. Regression from #1557 (v1.24), which moved the allocation into the loop;
the deferral was correct, the line just landed one block too deep.

The fix moves the (length-guarded) initialisation out of the loop, mirroring the
`wb$vml` init a few lines above.

Minimal reprex:

```r
library(openxlsx2)
img <- system.file("extdata", "einstein.jpg", package = "openxlsx2")

wb <- wb_workbook()
for (i in 1:3) {
  wb$add_worksheet()
  wb$add_comment(sheet = i, dims = "B2",
                 comment = wb_comment(text = "x", author = "x"), file = img)
}

count_vmlrels <- function(f)
  sum(grepl("vmlDrawing.*\\.vml\\.rels$", unzip(f, list = TRUE)$Name))

f1 <- temp_xlsx(); wb$save(f1)
f2 <- temp_xlsx(); wb_load(f1)$save(f2)
count_vmlrels(f1)  # 3
count_vmlrels(f2)  # before: 1, after: 3
```

* `R/wb_load.R` - init `wb$vml_rels` before the loop (length-guarded).
* `tests/testthat/test-loading_workbook.R` - regression test covering the
  three-`.vml.rels` round-trip.

The dangling `pageSetup r:id` (the other half of #1640 ) is left out on
purpose - it is a design question raised in the issue and not a regression.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Fixes
 
+* `wb_load()` no longer drops all but the last worksheet's `vmlDrawing*.vml.rels` on a load + save round-trip. The `wb$vml_rels` initialisation was inside the read loop and wiped on every iteration ([#1641](https://github.com/JanMarvin/openxlsx2/pull/1641), @SchmidtPaul).
+
 * Build vignette if `encharter` is not available.
 * Enhanced `df_to_xml()` speed and prevented double-escaping of XML entities, resolving an issue where hyperlinks with ampersands (`&`) were broken. [#1636](https://github.com/JanMarvin/openxlsx2/pull/1636), [#1637](https://github.com/JanMarvin/openxlsx2/pull/1637)
 * It is now possible to pass custom formula arguments to total rows when writing data tables. [#1638](https://github.com/JanMarvin/openxlsx2/pull/1638)
--- a/tests/testthat/test-loading_workbook.R
+++ b/tests/testthat/test-loading_workbook.R
@@ -742,3 +742,43 @@ test_that("has_drawings is defunct", {
     "argument \"has_drawing\" was removed"
   )
 })
+
+test_that("loading keeps every vmlDrawing.vml.rels (#1557 regression)", {
+
+  # A comment with a background image produces a vmlDrawing*.vml.rels per
+  # sheet. wb$vml_rels used to be re-initialised inside the load loop, so all
+  # but the last sheet's .vml.rels were dropped on a load + save round-trip,
+  # leaving dangling o:relid references (Excel repair prompt).
+  img <- system.file("extdata", "einstein.jpg", package = "openxlsx2")
+
+  wb <- wb_workbook()
+  for (i in 1:3) {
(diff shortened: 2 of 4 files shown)


tests/testthat/test-loading_workbook.R
```diff
@@ -742,3 +742,42 @@ test_that("has_drawings is defunct", {
...
+      sheet   = i,
+      dims    = "B2",
+      comment = wb_comment(text = paste("comment", i), author = "x"),
+      file    = img
+    )
+  }
+
+  count_vmlrels <- function(f)
+    sum(grepl("xl/drawings/_rels/vmlDrawing.*\\.vml\\.rels$",
+              unzip(f, list = TRUE)$Name))
```
JanMarvin, line 767: Could you please fix the `lintr` complain here.

NEWS.md
```diff
@@ -2,6 +2,8 @@
 
 ## Fixes
 
+* `wb_load()` no longer drops all but the last worksheet's `vmlDrawing*.vml.rels` on a load + save round-trip. The `wb$vml_rels` initialisation was inside the read loop and wiped on every iteration (regression in [#1557](https://github.com/JanMarvin/openxlsx2/pull/1557)).
```
JanMarvin: I don't think its necessary to link the regression here,  Just link this PR here like this:
`([1641](https://github.com/JanMarvin/openxlsx2/pull/1641), @SchmidtPaul)`.

JanMarvin (changes requested):
Could you please do the minor cleanups and please add yourself to the `AUTHORS` file here: `https://github.com/JanMarvin/openxlsx2/blob/main/inst/AUTHORS`.

Thanks!

tests/testthat/test-loading_workbook.R
SchmidtPaul: Done - wrapped the helper body in braces so `brace_linter` is happy.

NEWS.md
SchmidtPaul: Done, updated to `([#1641](https://github.com/JanMarvin/openxlsx2/pull/1641), @SchmidtPaul)`.

SchmidtPaul:
Added myself to `inst/AUTHORS`. All three points addressed and pushed - ready for another look, thanks!

JanMarvin (approved):

JanMarvin:
Thank you!

PR #1643, opened 2026-06-15
[load] drop dangling pageSetup printerSettings r:id on load

SchmidtPaul:
Fixes the `pageSetup r:id` half of #1640, following your preference in the issue
thread to remove the reference rather than ship the printer blob.

`wb_load()` keeps the worksheet/chartsheet `pageSetup` node verbatim, including
its `r:id`. That `r:id` only ever points at a `printerSettings*.bin` blob, which
openxlsx2 intentionally does not keep (the printerSettings relationship is
already dropped in the sheet rels handling). On `wb_save()` the `r:id` is written
back out, so it dangles and triggers Excel's repair prompt. Not a regression -
this has been the case since the rels filtering was introduced.

The fix strips the `r:id` from `pageSetup` right after the node is read, in both
the worksheet and the chartsheet load paths (both share the rels filtering that
removes printerSettings).

Minimal reprex (uses your `unemployment-nrw202208.xlsx`, which has 22 sheets each
carrying a `pageSetup r:id`):

```r
library(openxlsx2)
fl <- "unemployment-nrw202208.xlsx"  # from JanMarvin/openxlsx-data

count_rid <- function(f)
  sum(grepl("<pageSetup[^>]*r:id",
            vapply(grep("xl/worksheets/sheet", unzip(f, list = TRUE)$Name, value = TRUE),
                   function(s) paste(readLines(unz(f, s), warn = FALSE), collapse = ""),
                   character(1))))

f2 <- temp_xlsx(); wb_load(fl)$save(f2)
count_rid(fl)  # 22
count_rid(f2)  # before: 22 (dangling), after: 0
```

* `R/wb_load.R` - strip the `pageSetup` `r:id` on load (worksheet + chartsheet).
* `tests/testthat/test-loading_workbook.R` - round-trip regression test using
  `unemployment-nrw202208.xlsx`.

This branch is stacked on top of #1641, so the diff shown until #1641 is merged
includes that commit too; once #1641 lands this collapses to the single
pageSetup commit. Merging in order #1641 -> this PR is conflict-free.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Fixes
 
+* `wb_load()` no longer drops all but the last worksheet's `vmlDrawing*.vml.rels` on a load + save round-trip. The `wb$vml_rels` initialisation was inside the read loop and wiped on every iteration ([#1641](https://github.com/JanMarvin/openxlsx2/pull/1641), @SchmidtPaul).
+
 * Build vignette if `encharter` is not available.
 * Enhanced `df_to_xml()` speed and prevented double-escaping of XML entities, resolving an issue where hyperlinks with ampersands (`&`) were broken. [#1636](https://github.com/JanMarvin/openxlsx2/pull/1636), [#1637](https://github.com/JanMarvin/openxlsx2/pull/1637)
 * It is now possible to pass custom formula arguments to total rows when writing data tables. [#1638](https://github.com/JanMarvin/openxlsx2/pull/1638)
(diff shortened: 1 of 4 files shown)


JanMarvin:
Thanks! The file you picked, does not cause an issue for me on Mac with MS365. But different Excel versions behave differently. Could you please use something like this, this avoids me getting confused what the regexes is doing:

```R
pS <- wb$worksheets[[1]]$pageSetup
xml_attr_mod(pS, xml_attributes = c(`r:id` = ""))
```

SchmidtPaul:
Switched both load paths (worksheet + chartsheet) to your suggestion:

```r
ws$pageSetup <- xml_attr_mod(ws$pageSetup, xml_attributes = c(`r:id` = ""))
```

Much clearer, thanks. On the fixture: I kept `unemployment-nrw202208.xlsx` because the test asserts structurally that no `r:id` dangles after the round-trip, independent of how a given Excel build reacts. Happy to swap it if you'd prefer another file.

JanMarvin:
Thanks again! Sorry, busy day hadn't had much time, much appreciated!

Issue #1648, opened 2026-06-19
A few small independent fixes from a proactive bug hunt


SchmidtPaul:
I ran a small proactive bug hunt over openxlsx2 (AI assisted) and came out with three independent, verified fixes. Rather than dropping three PRs on you without context, here's a short overview - please take whichever ones you think are worthwhile and ignore the rest. Each lives on its own branch off `main` and can be merged independently, in any order.

1. **`[color]` `wb_color(name =, format = "RGBA")` returns the wrong colour.** `name` and `hex` get validated twice, so the RGBA alpha-swap is applied a second time - `name = "blue"` comes out red and that wrong colour is written into the file. Regression from #1341 (first shipped in v1.16). -> #1649
2. **`[args]` some functions silently absorb unknown arguments.** `wb_add_conditional_formatting()`, `wb_merge_cells()`, `wb_unmerge_cells()` and `wb_set_base_colors()` never call `standardize_case_names()`, so unknown/misspelled arguments are dropped with no warning at all - the silent-absorption case from #1646. -> #1650
3. **`[docs]` `wb_dims()` documents `rows = 0` for "column names only", but that always errors.** The documented behaviour was never implemented; `select = "col_names"` is the working way. Docs fix + a guarding test. -> #1651

All three are verified against a source build (fail-before / pass-after, relevant test files green). Happy to adjust, split, or drop any of them. I'll rebase the remaining PRs after each merge so the `NEWS.md` entries don't conflict.

JanMarvin:
Thank you, @SchmidtPaul !

I’ll look into them. I just removed the co-authored by Claude. I don’t like their aggressive advertising. I don’t co-author things to rstudio or visual studio code etc. Either the commit looks valid or not :)

SchmidtPaul:
Thanks. Yeah I get that - my intent was not to advertise, but to be transparent regarding AI usage.

JanMarvin:
Thanks, all merged

PR #1650, opened 2026-06-19
[args] warn on unknown arguments in CF, merge_cells, set_base_colors

SchmidtPaul:
Refs #1648

`wb_add_conditional_formatting()`, `wb_merge_cells()`, `wb_unmerge_cells()` and `wb_set_base_colors()` pull individual legacy keys (`cols`/`rows`, `xml`) out of `...` but never call `standardize_case_names()`, so any other argument is absorbed without a warning - unlike the `wb_add_*` functions. This is the silent-absorption case from #1646.

```r
library(openxlsx2)
wb <- wb_workbook()$add_worksheet()$add_data(x = mtcars[1:3, ])

wb$merge_cells(dims = "A1:B1", directon = "row")   # typo of 'direction'
#> before: (silent)   after: Warning: unused arguments (directon)
wb$set_base_colors(nonexistent = 1)
#> before: (silent)   after: Warning: unused arguments (nonexistent)
```

Fix: add the established `standardize_case_names(..., arguments = c(ls(), <legacy keys>))` guard at the top of each function (same pattern as `add_image()`). The deprecated `cols`/`rows`/`xml` keys keep working.

Scope note: R's partial matching already binds prefix typos (`solv`->`solve`, `styl`->`style`), so only non-prefix typos and genuinely unknown names were dropped silently - those now warn. `wb_set_last_modified_by()` is left out deliberately (its legacy key `LastModifiedBy` is camelCase and would be rewritten by `standardize_case_names()`, needing special handling) - happy to add it separately.

Changed: `R/class-workbook.R`, `tests/testthat/test-standardize.R`, `NEWS.md`. Verified against a source build (silent before / warn after; standardize + conditional_formatting + class-workbook tests green).


JanMarvin:
Thanks! I squeezed in a tiny commit of mine.

PR #1651, opened 2026-06-19
[docs] wb_dims(): rows = 0 never selected only the column names

SchmidtPaul:
Refs #1648

The `wb_dims()` documentation states: *"If `rows` = 0, only column names will be affected."* But `rows = 0` has always hit the `min(rows) < 1L` guard:

```r
wb_dims(x = mtcars, rows = 0)
#> Error: You must supply positive values to `rows`.
```

There has never been any `rows == 0` special-casing (since the line was added in #702), so the documented behaviour is unreachable. The working way to target only the header is `select = "col_names"`. This points the doc line there and adds a test pinning both the error and the working path. Docs only, no behaviour change.

Changed: `R/utils.R`, `man/wb_dims.Rd`, `tests/testthat/test-utils.R`, `NEWS.md`.


JanMarvin:
I assume we can roll with this, because nobody has complained so far. But chances are high that I broke something from @olivroy's initial implementation when I rewrote parts of it later on.

PR #1658, opened 2026-07-06
Add c2i()

tri-ton:
It adds `c2i()` as short interface to `col2int()` following [this discussion](https://github.com/ycphs/openxlsx/issues/273#issuecomment-951192885) to original openxlsx. I compared `col2int()` in openxlsx and openxlsx2 and see that the new version is significantly more advanced, so it doesn't make much sense to add this to openxlsx. Despite `c2i()` is a very short wrapper, I've also tried to add meaningful unittests.

I see the doubt about whitespaces as separators in the original discussion, but to my opinion they seem like the most short and readable option. They, of course, can be changed to commas or semicolumns if some style must be maintained.

### DIFF
--- a/R/converters.R
+++ b/R/converters.R
@@ -30,7 +30,7 @@
 #' * Non-integer numeric values will typically be coerced or truncated; however,
 #'     infinite values will trigger an error to prevent invalid coordinate
 #'     generation.
-#' @seealso [col2int()]
+#' @seealso [col2int()] for inverse operation
 #'
 #' @examples
 #' # Convert a single index
@@ -93,7 +93,9 @@ check_range <- function(x) {
 #' * In compliance with spreadsheet software standards, the function validates
 #'     that indices do not exceed the maximum allowable column limit.
 #'
-#' @seealso [int2col()]
+#' @seealso 
+#' [int2col()] for inverse operation
+#' [c2i()] for short interface
 #'
 #' @examples
 #' # Convert standard labels
--- a/R/converters.R
+++ b/R/converters.R
@@ -137,6 +137,43 @@ col2int <- function(x) {
   col_to_int(x)
 }
 
+#' Short way of converting spreadsheet column notations to integers
+#'
+#' @description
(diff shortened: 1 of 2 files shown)


JanMarvin:
Hi @tri-ton, thanks for reaching out and taking the time to open this PR.

While I appreciate the quality-of-life improvement this offers for your specific workflow, I would prefer not to add a new function to the package for this use case.

To keep the package maintainable and the API focused, I want to minimize adding functions that serve specific personal workflows. Since this function is tailored to your setup, it should live in a tiny helper script in your .Rprofile or in a local package. 

Thanks again for the contribution and for bringing back our discussion from a few years ago! As you mentioned, `openxlsx2` has evolved a lot in the last few years and now you can simply run:

```R
c("B:G","X","AF:AK","BH") |>
  openxlsx2::col2int()
```

tri-ton:
Hi @JanMarvin, no problem and thanks a lot for the quick response! Actually, I already had similar function in my personal workflow for quite some time; I've opened the PR because some of my less experienced colleagues asked about this. Would you suggest trying to open a similar PR in original `openxlsx`? As far as I understood, new functions can be accepted there if fully implemented. If that's not an option, I'll just suggest the colleagues to use `readxl` (it is usually less convenient for new users than `openxlsx`/`openxlsx2`).

JanMarvin:
Hi @tri-ton , thanks for understanding. For `openxlsx` I'd say the same. I would accept a PR that brings `col2int()` on par with the `openxlsx2` implementation, but I don't think its a good idea to give people the idea that there is new development in `openxlsx`.

I'm still not convinced that its a good idea to allow it, but I have added a PR that splits col2int input on "," and ";" on single string input. So the following should be possible:

```R
c("B:G,X,AF:AK,BH") |>
  openxlsx2::col2int()
```

I assume your use case is something like this?

```R
# previous
openxlsx2::wb_to_df(datasets, cols = c("A","C:D"))
# with this PR
openxlsx2::wb_to_df(datasets, cols = "A,C:D")
```

tri-ton:
That's exactly the use case. Thanks a lot!

PR #1669, opened 2026-08-04
[misc] Update print method for workbook objects

felix-hof:
Hi,

first of all thank you for maintaining this great package! I've been using it a lot lately. However, what bugged me a bit is that the print method for workbook objects does not print a newline character at the end. So while the current method generates this output:

```r
>   wb <- wb_workbook()$
+     add_worksheet("Sheet 1")$
+     add_worksheet("Sheet 1 (1)")$
+     add_worksheet("Sheet & NoSheet")
> print(wb)
A Workbook object.
 
Worksheets:
 Sheets: Sheet 1, Sheet 1 (1), Sheet & NoSheet 
 Write order: 1, 2, 3> 
```

the changes in this PR just add a newline to the end (Note how the prompt `>` is set to the next line).

```r
> print(wb)
A Workbook object.
 
Worksheets:
 Sheets: Sheet 1, Sheet 1 (1), Sheet & NoSheet 
 Write order: 1, 2, 3 
> 
```

### DIFF
--- a/R/class-workbook.R
+++ b/R/class-workbook.R
@@ -7421,7 +7421,7 @@ wbWorkbook <- R6::R6Class(
           ))
       }
 
-      cat(unlist(showText))
+      cat(unlist(showText), "\n")
       invisible(self)
     },
 
--- a/inst/AUTHORS
+++ b/inst/AUTHORS
@@ -12,6 +12,7 @@ David Gohel
 david-f1976
 debarros
 Evan Cortens
+Felix Ventura de Oliveira
 Florian Weimer
 Gregory Fiumara
 Gregory Warnes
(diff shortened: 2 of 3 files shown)


JanMarvin:
Hi @felix-hof , 

thank you for your kind words!

Regarding the PR, I see that our print behaves differently in Rstudio vs the native console. In Rstudio it already looks fine, but in a terminal the behavior you describe is visible. Can we make it work for both? I would live to avoid creating additional blank lines for Rstudio users (aka myself).

JanMarvin:
Oh, it does not matter. Works for Rstudio just fine. Can you please add yourself to [inst/AUTHORS](https://github.com/JanMarvin/openxlsx2/blob/main/inst/AUTHORS) (either full name or github handle).

felix-hof:
Hi @JanMarvin,

Sorry, I didn't even think of testing this in RStudio. Anyway, thanks for trying it! I also added my name to the AUTHORS file.

JanMarvin:
Thanks, merged!

PR #1684, opened 2026-09-08
[cf] fix rule leaking from block to block with non consecutive dims

SchmidtPaul:
`wb_add_conditional_formatting()` still writes wrong rules for a non consecutive `dims`, which is the complaint of #1346 ("I ended up with broken xlsx instead of failing quickly"). #1347 [split `dims` into consecutive row and column blocks and looped over them](https://github.com/JanMarvin/openxlsx2/blob/6a391297c375ca1021d7bf73bf64e6f98cd12994/R/class-workbook.R#L6172-L6183) and does solve the `expression` case that #1346 actually reported, but only that one branch inside the loop was made re-entrant. For every other `type` the second and any further block is built from what the previous iteration left behind.

I ran into it while porting the Excel helpers of my own package from openxlsx to openxlsx2: a color scale over a set of rows with gaps renders as a scale for the first block and flat for the rest.

Six branches read `rule` into `values` and then overwrite it with `style` ([`colorScale`](https://github.com/JanMarvin/openxlsx2/blob/6a391297c375ca1021d7bf73bf64e6f98cd12994/R/class-workbook.R#L6238-L6239), `dataBar`, `containsText`, `notContainsText`, `beginsWith`, `endsWith`), and only `expression` restores it from [`orig_rule`](https://github.com/JanMarvin/openxlsx2/blob/6a391297c375ca1021d7bf73bf64e6f98cd12994/R/class-workbook.R#L6180). The other branches that assign `rule <- style` never read `rule`, so for them it is idempotent. Gaps in rows and gaps in columns both trigger it. Introduced with #1347 (v1.16), which is where the loop came in; before that a non consecutive `dims` was reduced to its bounding box.

```r
library(openxlsx2)

wb <- wb_workbook()$add_worksheet()$add_data(x = data.frame(v = 1:5))
wb$add_conditional_formatting(
  dims  = wb_dims(rows = c(2, 3, 5), cols = 1),
  type  = "colorScale",
  style = c("#FFFFFF", "#63BE7B")
)
wb$worksheets[[1]]$conditionalFormatting$cf[2]
#> main:    <cfRule type="colorScale" priority="2"><colorScale><cfvo type="num" val="FFFFFFFF"/><cfvo type="num" val="FF63BE7B"/><color rgb="FFFFFFFF"/><color rgb="FF63BE7B"/></colorScale></cfRule>
#> this PR: <cfRule type="colorScale" priority="2"><colorScale><cfvo type="min"/><cfvo type="max"/><color rgb="FFFFFFFF"/><color rgb="FF63BE7B"/></colorScale></cfRule>

wb <- wb_workbook()$add_worksheet()$add_data(x = data.frame(a = "b", b = "b", c = "b"))
wb$add_dxfs_style(name = "hit", bg_fill = wb_color("red"))
wb$add_conditional_formatting(
  dims  = wb_dims(rows = 2, cols = c(1, 3)),
  type  = "containsText",
  rule  = "b",
  style = "hit"
)
wb$worksheets[[1]]$conditionalFormatting$cf[2]
#> main:    <cfRule type="containsText" dxfId="0" priority="2" operator="containsText" text="hit"><formula>NOT(ISERROR(SEARCH("hit", C2)))</formula></cfRule>
#> this PR: <cfRule type="containsText" dxfId="0" priority="2" operator="containsText" text="b"><formula>NOT(ISERROR(SEARCH("b", C2)))</formula></cfRule>

wb <- wb_workbook()$add_worksheet()$add_data(x = data.frame(v = 1:5))
wb$add_conditional_formatting(
  dims  = wb_dims(rows = c(2, 3, 5), cols = 1),
  type  = "dataBar",
  style = "#63BE7B"
)
#> main:    Error in values[[2]] : subscript out of bounds
wb$worksheets[[1]]$conditionalFormatting$cf[2]
#> this PR: <cfRule type="dataBar" priority="2"><dataBar showValue="1"><cfvo type="min"/><cfvo type="max"/><color rgb="FF63BE7B"/></dataBar>...
```

The fix moves the reset out of the `expression` branch to the top of the loop, so every block starts from what the user passed. The one behaviour change beyond that: with an invalid `rule` the per block validation now runs per block, so `type = "between"` with `rule = NULL` warns four times instead of two.

The `switch()` is loop invariant except for the cell reference that `expression` builds, so hoisting it out of the loop would make this class of leak structurally impossible. That is a refactor of a 200 line function though, so I kept it out of a fix that should be safe for a patch release. Happy to open it separately if you want it.

Not affected, and left alone: `between` (`range()` is idempotent), `topN`/`bottomN` (their values come from `params`), `iconSet` (never reassigns `rule`), and the `*Blanks`/`*Errors`/`duplicatedValues`/`uniqueValues` branches. I ran all 17 types against both builds: the six above are the only ones whose second block changes, and the first block is byte identical for every one of them, so nothing changes for a consecutive `dims`. The block loop is also the only one of its kind here - `merge_cells()` got its non consecutive `dims` in #1357 by splitting the `dims` string and treating each piece on its own, so it has no per block state to leak.

Changed files:

* `R/class-workbook.R` - reset `rule` from `orig_rule` at the top of the block loop instead of inside the `expression` branch
* `tests/testthat/test-conditional_formatting.R` - one regression test per affected family (`colorScale` including an explicit `rule` and a four block `dims`, `dataBar`, the four text types), plus an `expression` case, which passes on `main` and is there to pin the reset that used to live in that branch
* `NEWS.md`

Verified against real source builds of `main` and this branch: on `main` the new tests produce 8 failures and 1 error, on this branch `test-conditional_formatting.R` is green, and `R CMD check` shows the same result as `main` (one unrelated pre-existing test failure over a missing `testfiles/loadExample.xlsx`).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@
 ## Fixes
 
 * A logic bug in `apply_numfmts()` was fixed that occured in combination with `cols` used via `wb_set_col_widths()`. In addition automatic sizing can now be tweaked via an offset `"auto+n"`. [#1683](https://github.com/JanMarvin/openxlsx2/pull/1683)
-* `wb_add_conditional_formatting()` no longer builds every block but the first from the previous block's leftovers when `dims` is non consecutive. `rule` was overwritten inside the loop over the blocks, so a `"colorScale"` got the colors as `cfvo` values, a `"containsText"` searched for the style name and a `"dataBar"` failed with `subscript out of bounds` (broken for every type but `"expression"` since [#1347](https://github.com/JanMarvin/openxlsx2/pull/1347), @SchmidtPaul).
+* `wb_add_conditional_formatting()` no longer builds every block but the first from the previous block's leftovers when `dims` is non consecutive. `rule` was overwritten inside the loop over the blocks, so a `"colorScale"` got the colors as `cfvo` values, a `"containsText"` searched for the style name and a `"dataBar"` failed with `subscript out of bounds`. Broken for every type but `"expression"` since [#1347](https://github.com/JanMarvin/openxlsx2/pull/1347) ([#1684](https://github.com/JanMarvin/openxlsx2/pull/1684), @SchmidtPaul).
 
 
 ***************************************************************************
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 ## Fixes
 
 * A logic bug in `apply_numfmts()` was fixed that occured in combination with `cols` used via `wb_set_col_widths()`. In addition automatic sizing can now be tweaked via an offset `"auto+n"`. [#1683](https://github.com/JanMarvin/openxlsx2/pull/1683)
+* `wb_add_conditional_formatting()` no longer builds every block but the first from the previous block's leftovers when `dims` is non consecutive. `rule` was overwritten inside the loop over the blocks, so a `"colorScale"` got the colors as `cfvo` values, a `"containsText"` searched for the style name and a `"dataBar"` failed with `subscript out of bounds` (broken for every type but `"expression"` since [#1347](https://github.com/JanMarvin/openxlsx2/pull/1347), @SchmidtPaul).
 
 
 ***************************************************************************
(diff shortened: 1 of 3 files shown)


JanMarvin:
Thank you, @SchmidtPaul ! My wording would have been a bit more on the "complete `openxlsx2` support for non consecutive CF", because its been more than year by now and nobody complained so far. But I might be biased 😄

PR #1689, opened 2026-09-21
[cf] apply a single rule to a non consecutive dims

### LINKED ISSUE #1688: [cf] complete support for non consecutive dims
The same `dims` reaches `wb_add_fill()` and `wb_add_conditional_formatting()` differently:

```r
wb <- wb_workbook()$add_worksheet()$add_data(x = matrix(1, 5, 5), col_names = FALSE)

wb$add_fill(dims = "A1:B2,C2:D3", color = wb_color("red"))
# fills the 8 cells that were named

wb$add_conditional_formatting(dims = "A1:B2,C2:D3", rule = "==1")
wb$worksheets[[1]]$conditionalFormatting$sqref
#> "A1:D3"     <- 12 cells, 8 were named
```

`"A1:B2,C2:D3"` is the example `NEWS.md` gives for the non consecutive `dims` from #1347. Losing cells out of such a `dims` is what #1014 reported for `add_fill()` and #1015 fixed; the conditional formatting path pools them instead. #469 already states the goal: "If dims `"A1:D1;B2:E2"` is requested, only these cells should be filled."

```
dims = "A1:B2,C2:D3"        # formatted   . untouched

is                     should
   A B C D                A B C D
1  # # # #             1  # # . .
2  # # # #             2  # # # #
3  # # # #             3  . . # #
```

The rows and columns of all blocks are pooled ([`class-workbook.R#L6140-L6143`](https://github.com/JanMarvin/openxlsx2/blob/b13f844d9c6e7767bfd2cf63947690e1574cbd82/R/class-workbook.R#L6140-L6143)) and the loop then runs over their cross product ([`#L6172-L6183`](https://github.com/JanMarvin/openxlsx2/blob/b13f844d9c6e7767bfd2cf63947690e1574cbd82/R/class-workbook.R#L6172-L6183)), which equals the selection only when the blocks fill that cross product.

**A first step** would be to keep the blocks apart and merge them back only along an axis they agree on: same columns and touching rows, or same rows and any columns. That gives `"A1:B2" "C2:D3"` and stays inside that one loop. Selections the pooling already got right come out byte identical - I checked all 17 rule types against 9 such shapes, including rectangles written as several blocks like `"A1:B4,C1:C4"` - and it is faster rather than slower (three blocks over 200k rows: 0.22 s to 0.13 s here). I have it on a branch with tests if you want it.

**It would not be the whole story though.** Each block becomes its own `conditionalFormatting` element, so its own `cfRule`. For the types that aggregate over their range (`colorScale`, `dataBar`, `iconSet`, `topN`/`bottomN`, `duplicatedValues`, `uniqueValues`) that means one scale per block rather than one over the selection: a `colorScale` over an L shape ends up with two independent gradients. And merging can only get so far. `"A1:B1,C1:C2,B3:C3,A2:A3,B2"` is the rectangle A1:C3 tiled like a windmill, where no two blocks share a full edge, so it stays five elements where pooling produced one.

**The complete version** would be a single element carrying all ranges, `sqref="A1:B2 C2:D3"`, which OOXML allows. That would make the decomposition irrelevant: however `dims` is written, the selection gets one rule. Most of the machinery is there already - the writer emits whatever string sits in `sqref`, and when I collapse the elements by hand and save, such a file round trips through `wb_load()` unchanged. What cannot express it is `do_conditional_formatting()`, which joins exactly one range out of `startRow`/`endRow`/`startCol`/`endCol` ([`#L10545-L10548`](https://github.com/JanMarvin/openxlsx2/blob/b13f844d9c6e7767bfd2cf63947690e1574cbd82/R/class-workbook.R#L10545-L10548)). `dataBar` and `iconSet` carry their own `sqref` in the `x14` extension and would need the same.

The first step would not be thrown away by that: the grouping is what turns `dims` into the list of ranges a multi range `sqref` needs.

I have only checked that such a file round trips through openxlsx2, not how Excel renders a multi range `sqref`.

Would you rather have the first step now, or the `sqref` version in one go? And for the aggregating types, would one `colorScale` over the whole selection be what you want there, or is one per block defensible?

SchmidtPaul:
Closes #1688, refs #469

`wb_add_conditional_formatting()` now writes one `conditionalFormatting` element for a non consecutive `dims`, carrying every range in its `sqref`:

```r
library(openxlsx2)

wb <- wb_workbook()$add_worksheet()$add_data(x = matrix(1, 5, 5), col_names = FALSE)
wb$add_conditional_formatting(dims = "A1:B2,C2:D3", rule = "==1")
wb$worksheets[[1]]$conditionalFormatting$sqref
#> master:  "A1:D3"          12 cells, 8 were named
#> this PR: "A1:B2 C2:D3"
```

That is what #469 asks for ("If dims `"A1:D1;B2:E2"` is requested, only these cells should be filled.") and it is what Excel writes for such a selection. Rows and columns used to be pooled across all blocks and the loop ran over their cross product ([`class-workbook.R#L6140-L6183`](https://github.com/JanMarvin/openxlsx2/blob/b13f844d9c6e7767bfd2cf63947690e1574cbd82/R/class-workbook.R#L6140-L6183)), which equals the selection only when the blocks fill that cross product, and every block re-anchored the rule.

**What Excel does.** Checked through COM against Excel 16.0, for every rule type I could construct there: `expression`, `colorScale`, `dataBar` (including its `<xm:sqref>`), `iconSet`, `containsText`, `containsErrors`, `containsBlanks`, `top10`, `duplicateValues`, `cellIs`/`between` and `beginsWith`. Excel writes a single element with a space separated `sqref` for all of them, anchors relative references to the top left cell of the first range, and keeps that `sqref` and the formula unchanged when it re-saves the file. `wb_load()` already reads such an element back unchanged.

**Blocks that overlap.** A `sqref` must not name a cell twice, because Excel counts it twice. Over `A1:C3 B2:D4` on a sheet where all 25 values differ, a `duplicatedValues` rule marks B2:C3 as duplicates of themselves and `uniqueValues` drops those four cells (read back through `DisplayFormat`). Excel cuts the overlap away instead: `"A1:C3,B2:D4"` becomes `A1:C3 B4:C4 D2:D4`, `"A1:D4,B2:C3"` becomes `A1:D4` and `"A1:B2,A1:B2"` becomes `A1:B2`. This PR cuts the same cells away, and cuts the way Excel does, columns first: what lies beside an earlier block keeps the full height of its own block, and only in the shared columns is what lies above and below cut off. It writes the ranges top to bottom and left to right rather than in the order they were cut, so the first of those is `A1:C3 D2:D4 B4:C4` here, the same three ranges as Excel's. Blocks are sorted before they are cut, so the ranges do not depend on the order the blocks were written in and `wb_remove_conditional_formatting()` still finds them. Against 20 random overlapping selections the result covers the same cells as Excel's, to the cell count, with no cell listed twice; the same holds over 4000 random selections against the union computed by hand.

**Your moving column example.** `A$1<0` over `"A1:B6,B7:C9"` with row 1 = `-1, 1, -1, 1` colours column A, leaves column B, colours C7:C9 and leaves the unselected cells of column C and B7. So C8 checks `C$1`, as the chapter promises, and only the 18 named cells are touched. A `colorScale` over `"A1:B2,C2:D3"` is one gradient from A1 to D3 with C1 and A3 untouched. The files open without a repair dialog and `AppliesTo` is the selection. Nothing in the book chapter or in the vignettes uses a non consecutive `dims` ("Moving Col" is `"A1:B11"`), so no rendered example changes.

Changes:

* a new internal `cf_dims_to_sqref()` keeps the blocks of `dims` apart and merges them back only along an axis they agree on (same columns and touching rows, or same rows), to a fixed point. The result is the range list of the `sqref`. It runs before a style is registered, so an unusable `dims` now leaves the workbook untouched instead of leaving a dxfs entry behind, and an empty selection is an error
* a second internal `cf_cut_overlaps()` removes an overlap between blocks before that, so no cell is named twice. Blocks whose bounding boxes do not meet skip it for the price of one matrix, so a `dims` of many disjoint blocks is unaffected
* the ranges are written top to bottom, left to right, whatever order the blocks were given in, so the anchor does not depend on the spelling. Excel keeps the order of the selection instead; for a `dims` that is already in that order, the normal case, the two agree
* `do_conditional_formatting()` takes the `sqref` instead of one `startRow/endRow/startCol/endCol`; the text, error and blank rules keep writing the first range into their formula, `dataBar` and the four `iconSet` variants that live in the x14 extension carry the full list in `xm:sqref` as Excel does
* `wb_remove_conditional_formatting()` goes through the same range builder, so it now finds the stored element whichever way the merge normalises the same selection: `"A1:B4,C1:C4"` as well as `"A1:C4"`, and `"A3:B4;C1:D2"` as well as `"C1:D2,A3:B4"`. A decomposition that agrees on neither axis stays several ranges and does not match. A `dims` vector keeps being matched element by element, as it was when this compared against the stored `sqref` directly
* the details paragraph that said the expression "will be evaluated again as `A1=1`" for the second range is replaced

**Compatibility.** Single range output is unchanged: the `sqref` and the `cfRule` XML are identical to master for all 17 rule types against six rectangular selections, including these rectangles written as several blocks (`"A1:B4,C1:C4"`, `"A1:A4,B1:B4,C1:C4"`, `"A1:B2,C1:D2,A3:B4,C3:D4"`). A pinwheel tiling of a rectangle shares no axis pairwise and stays five ranges over the same cells, one element either way.

On not breaking anything: I ran `correlation::cormatrix_to_excel()` (behind `rempsyc::cormatrix_excel()`) on a synthetic five column frame against source builds of master and this branch. Both workbooks are byte identical, every part except `docProps`, with the same 20 `conditionalFormatting` elements at `sqref="B2:F6"`. It passes a single rectangle, so it never enters the changed path; a test in that shape is included. What does change is any call that already passed a non consecutive `dims`: it gets one element instead of one per block, so a `colorScale` over `"A1:A11,C1:C11"` becomes one scale over both columns. The tests from #1684 that pinned one element per block are rewritten accordingly, and NEWS lists this under breaking changes.

Changed files:

* `R/conditional_formatting.R` - the new `cf_dims_to_sqref()`, `cf_cut_overlaps()` and `cf_block_minus()`
* `R/class-workbook.R` - `add_conditional_formatting()` and `remove_conditional_formatting()` use it, `do_conditional_formatting()` takes the `sqref`
* `R/class-workbook-wrappers.R` and `man/wb_add_conditional_formatting.Rd` - the details paragraph (the `.Rd` line edited by hand to match, no `roxygenise()` run)
* `tests/testthat/test-conditional_formatting.R`
* `NEWS.md`

Verified against real source builds of master and this branch in an isolated library: `test-conditional_formatting.R` 162 passed and 0 failed (39 failed and 1 error on master), full suite 2491 passed, 0 failed and 10 skipped, lintr clean on the changed files. A `colorScale` over three blocks of 200k rows goes from 0.20 s to 0.12 s (medians of 7, three fresh processes each). The one case that gets slower is a `dims` of 1000 blocks that all overlap, 0.09 s to 0.11 s; 1000 disjoint blocks stay at 0.22 s against 1.03 s on master.

The diff of `class-workbook.R` is mostly indentation from the two loops that are gone.

*This contribution was prepared with the assistance of an AI coding tool.*


JanMarvin:
Thank you! Had no time to review this today, but a quick look was fine. Really appreciate you checking it with Excel. That’s the tedious stuff that takes so much time. Unfortunately that’s the only way to know if the implementation is fine.

JanMarvin:
@SchmidtPaul , thanks, can you review the commit I added afterwards? I'm a bit unhappy with the amount of new code we will introduce with this commit, but if that is needed, so be it. And its touching only conditional formatting. I trust you with that and I can live with that.

SchmidtPaul:
Thanks, your three changes are all improvements! And you are right about the amount of code, so I went back into it with the goal of shortening it. While doing that I found two small things in the cleanup commit: passing the sqref of a loaded file verbatim (`"A1:C3 D2:D4 B4:C4"`) to `wb_remove_conditional_formatting()` errored, and a stored sqref the builder cannot read, like `"1:1"`, stopped the removal of every other rule. Both fixed and tested. Given and family in `person()` were also swapped.

However, while I appreciate the trust, I ask you to please not merge based on trust alone. As the description says, I did rely on an AI coding tool, and ultimately I am not the expert I would like to be about some aspects of this. I did check, validate and add tests, including against Excel itself, but you are the one who will have to stand behind this code, so please only take it if you do. If it is still more than the feature is worth to you, closing this is fine with me as well.

JanMarvin:
Well, none of us was born an expert. The only way to get closer to becoming one is by making many mistakes and learning from these mistakes. But I will run a few more tests.

Sorry for the name switching, lower than average level of coffee in my system.

